PR #64, opened 2024-12-18
Improve relayer run scripts/docker

obrok:


README.md
```diff
@@ -70,6 +70,18 @@ Compile & deploy the smart contract suite:
 make deploy-contracts
 ```
 
+### Running e2e tests
+
+Run:
+
+```bash
+run tooling-e2e-tests/full_scenario.sh
```
pmikolajczyk41, line 78: let's maybe add the `make deps` and `make generate-contracts` commands that are commonly forgotten

crates/shielder-relayer/run-relayer.sh
```diff
@@ -18,7 +18,7 @@ REQUIRED_RUN_VARS=(
 )
 
 for var in "${REQUIRED_RUN_VARS[@]}"; do
-  if [ -z "${var}" ]; then
+  if [ -z "${!var+set}" ]; then
```
pmikolajczyk41, line 21: do we have any indirection in our scripts? or is this just more making things more robust?

pmikolajczyk41 (approved):

crates/shielder-relayer/run-relayer.sh
obrok: I don't quite understand the question... From what I understand the previous version checked if the var `var` was set (and it was to the name of the var to check), while this version checks if the var with the name in `var` is set.

obrok: >"Or else it doesn't, you know. The name of the song is called 'Haddocks' Eyes'."
"Oh, that's the name of the song, is it?" Alice said, trying to feel interested.
"No, you don't understand," the Knight said, looking a little vexed.
"That's what the name is called. The name really is 'The Aged Aged Man'."
"Then I ought to have said 'That's what the song is called'?" Alice corrected herself.
"No, you oughtn't: that's quite another thing! The song is called 'Ways And Means': but that's only what it's called, you know!"
"Well, what is the song, then?" said Alice, who was by this time completely bewildered.
"I was coming to that," the Knight said. "The song really is 'A-sitting On A Gate': and the tune's my own invention."

PR #71, opened 2024-12-23
Check if inputs are from field

pmikolajczyk41:
Add checks in the contract whether inputs are valid field elements. Accompanied with tests.

### DIFF
--- a/contracts/Shielder.sol
+++ b/contracts/Shielder.sol
@@ -144,10 +144,10 @@ contract Shielder is
         whenNotPaused
         withinDepositLimit
         restrictContractVersion(expectedContractVersion)
+        fieldElement(newNote)
+        fieldElement(idHash)
     {
         uint256 amount = msg.value;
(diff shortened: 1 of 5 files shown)


contracts/Shielder.sol
```diff
@@ -182,6 +189,12 @@ contract Shielder is
     {
         uint256 amount = msg.value;
         if (amount == 0) revert ZeroAmount();
+        if (
+            oldNullifierHash >= FIELD_MODULUS ||
+            newNote >= FIELD_MODULUS ||
+            idHiding >= FIELD_MODULUS
+        ) revert NotAFieldElement();
+
```
krzysztofziobro, line 197: Would try to wrap the check into a modifier/function

pmikolajczyk41: I used modifier, thanks for the idea!

krzysztofziobro (approved):

PR #76, opened 2025-01-06
ZK-643: General bindings crate

kroist:
This pull request introduces a new crate, `shielder_bindings`, which provides WebAssembly (WASM) and mobile bindings (uniffi-rs) for the Shielder cryptographic operations.

Key changes include:

### New Crate Setup:
* [`crates/shielder_bindings/Cargo.toml`]: Added new crate `shielder_bindings` which is mostly a copy of `shielder-wasm` crate, but currently only with poseidon hash and private-key-to-scalar conversion.

### Build and Configuration:
* [`crates/shielder_bindings/Makefile`]: Added separate makefile targets for building WASM packages and iOS bindings.
* [`crates/shielder_bindings/Cargo.toml`]: Added feature flags for binding generation options:
    - `build-wasm` - webassembly 
    - `multithreading-wasm` - optional multithreaded webassembly
    - `build-uniffi` - mobile bindings through uniffi

### Documentation:
* [`crates/shielder_bindings/README.md`]: Added README with an overview, build instructions for WebAssembly and iOS, and prerequisites.

PR #77, opened 2025-01-06
ZK-559: Add shielder-sdk-crypto package

kroist:
Introducing a typescript package `shielder-sdk-crypto` which purpose is to define an interface for platform-native cryptographic functions.

### Initial Setup and Configuration:

* Added `shielder-sdk-crypto` package to the workspace (`ts/pnpm-workspace.yaml`). We will eventually transfer all typescript packages in monorepo to the workspace.

### Core Functionality Implementation:

* Defined TypeScript interfaces and types for inputs and outputs of cryptographic operations (`ts/shielder-sdk-crypto/src/index.ts`).
* Implemented `CryptoClient` interface with methods for calling platform-native cryptography functions (`ts/shielder-sdk-crypto/src/cryptoClient.ts`).
* Added `Scalar` class and utility functions for handling scalar values. It is cross-platform vanilla TS/JS code (`ts/shielder-sdk-crypto/src/scalar.ts`).


ts/shielder-sdk-crypto/eslint.config.mjs
```diff
@@ -0,0 +1,40 @@
+import eslint from "@eslint/js";
```
pmikolajczyk41, line 1: can we share these config files among workspace members?

ts/shielder-sdk-crypto/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export type Proof = Uint8Array;
+
+export interface NewAccountPubInputs {
+  hNote: Scalar;
+  hId: Scalar;
+  initialDeposit: Scalar;
+}
+
+export interface NewAccountValues {
```
pmikolajczyk41, line 12: suggestion: rename `Values` to `Inputs`, `PrivateInputs`, `ProverKnowledge`, `Advice`

ts/shielder-sdk-crypto/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  trapdoorOld: Scalar;
+  accountBalanceOld: Scalar;
+  path: Uint8Array;
+  value: Scalar;
+  nullifierNew: Scalar;
+  trapdoorNew: Scalar;
+  commitment: Scalar;
+}
+
+export type ShielderActionSecrets = {
```
pmikolajczyk41, line 61: do we want id here as well?

ts/shielder-sdk-crypto/src/cryptoClient.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  proveDeposit(values: DepositValues): Promise<Proof>;
+
+  verifyDeposit(proof: Proof, pubInputs: DepositPubInputs): Promise<boolean>;
+
+  proveWithdraw(values: WithdrawValues): Promise<Proof>;
+
+  verifyWithdraw(proof: Proof, pubInputs: WithdrawPubInputs): Promise<boolean>;
+
+  proveAndVerifyMerkle(): Promise<Proof>;
```
pmikolajczyk41, line 29: do we still care about merkle?

ts/shielder-sdk-crypto/src/cryptoClient.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  poseidonHash(inputs: Scalar[]): Promise<Scalar>;
+
+  getSecrets(id: Scalar, nonce: number): Promise<ShielderActionSecrets>;
+
+  privateKeyToScalar(hex: `0x${string}`): Promise<Scalar>;
+
+  treeHeight(): Promise<number>;
```
pmikolajczyk41, line 37: I suggest using 100% clear name, like `noteTreeHeight` - in some time we might have other trees

pmikolajczyk41: similar comment to `arity` - it's not clear if that's for note tree, poseidon width or e.g. some token operation

ts/shielder-sdk-crypto/src/cryptoClient.ts
```diff
@@ -0,0 +1,40 @@
...
+  NewAccountPubInputs,
+  NewAccountValues,
+  Proof,
+  Scalar,
+  ShielderActionSecrets,
+  WithdrawPubInputs,
+  WithdrawValues
+} from "./index";
+
+export interface CryptoClient {
```
pmikolajczyk41, line 13: I don't know if that's feasible in ts, but if possible, I would split this interface - currently it mixes a lot of responsibilities:
 - proving
 - veryifying
 - configuration information (arity, height)
 - crypto primitives (hash)
 - conversion
 - secret management

I'm not saying that each of this bullet point shall be under a dedicated interface, but still, some separation might be useful and clarifying

ts/shielder-sdk-crypto/src/scalar.ts
```diff
@@ -0,0 +1,102 @@
+/**
```
pmikolajczyk41, line 1: the same comment as in #76 please flag somehow (e.g. by deprecation marker or in Jira) that there is some code duplication that needs to be removed as soon as possible (to avoid any divergence)

pmikolajczyk41 (approved):

ts/shielder-sdk-crypto/eslint.config.mjs
kroist: moved to workspace, it seems to work.

ts/shielder-sdk-crypto/src/index.ts
kroist: renamed to `*Advice`

ts/shielder-sdk-crypto/src/index.ts
kroist: id and nonce are parameters of the function, keeping it simple for now, no "mother-types".

ts/shielder-sdk-crypto/src/cryptoClient.ts
kroist: removed

ts/shielder-sdk-crypto/src/cryptoClient.ts
kroist: moved to `NoteTreeConfig` interface

ts/shielder-sdk-crypto/src/cryptoClient.ts
kroist: did split it into sub-interfaces

guspiel: If it were pure TypeScript, I would agree. But the purpose of `CryptoClient` is to aggregate the platform-specific functionality. Sometimes we'll inject a WASM crypto client, sometimes a React Native crypto client, and sometimes a mock. It feels simpler to do it once. Also, having just one gateway gives a single entry point for platform-specific initialization, if such initialization were needed. I think it might be better to keep these under 1 interface.

guspiel: Ok, now it is ok. Max just grouped the new interfaces inside CryptoClient and I am satisfied with that.

ts/shielder-sdk-crypto/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "name": "shielder-sdk-crypto",
+  "version": "0.0.1",
+  "description": "",
+  "scripts": {
+    "build": "tsc --project tsconfig.json && tsc-alias -p tsconfig.json",
+    "lint": "eslint . --report-unused-disable-directives --max-warnings 0"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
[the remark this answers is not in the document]
kroist: fixed to apache

ts/shielder-sdk-crypto/src/cryptoClient.ts
guspiel: @kroist just a gentle reminder to leave a short response if you decide not to apply a suggestion

PR #82, opened 2025-01-10
ZK-664: Extract type conversions crate

pmikolajczyk41:


kroist:
good, something is broken in `shielder-wasm` compilation but LGTM.

is this the only split or there will be more?

pmikolajczyk41:
@kroist no, there will be follow ups; but I wanted to go with baby steps

PR #86, opened 2025-01-14
ZK-1337: add "@cardinal-cryptography" prefix to ts packages and correct versions

kroist:


ts/shielder-sdk-crypto-wasm/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "shielder-sdk-crypto-wasm",
-  "version": "1.0.0",
+  "name": "@cardinal-cryptography/shielder-sdk-crypto-wasm",
+  "version": "0.2.0",
```
pmikolajczyk41, line 3: do we have to bump?

pmikolajczyk41 (approved):

ts/shielder-sdk-crypto-wasm/package.json
kroist: it's not released yet, `1.0.0` version was generated automatically

PR #87, opened 2025-01-14
ZK-1337: Publish packages to npm

kroist:
- removed workflow for publishing dev-env (nobody uses these scripts)
- instead added workflow for publishing npm packages

example workflows (actually pushed to npm):
- https://github.com/Cardinal-Cryptography/zkOS-monorepo/actions/runs/12772694131/job/35602855142
- https://github.com/Cardinal-Cryptography/zkOS-monorepo/actions/runs/12772787873/job/35603173157
- https://github.com/Cardinal-Cryptography/zkOS-monorepo/actions/runs/12772893812/job/35603521314

### DIFF
--- a/.github/workflows/_ts-checks.yml
+++ b/.github/workflows/_ts-checks.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 23
 
       - uses: pnpm/action-setup@v4
         with:
@@ -73,3 +73,20 @@ jobs:
       - name: Run lint checks (workspace)
         run: pnpm lint
         working-directory: ts
+
+      - name: Run ts tests (workspace)
+        run: pnpm test
+        working-directory: ts
+
+      - name: Generate coverage data
+        run: pnpm coverage-report
+        working-directory: ts/shielder-sdk
+
+      - name: Generate coverage markdown report
+        run: node __tests/generate-coverage-report.js
+        working-directory: ts/shielder-sdk
+
+      - name: Post coverage report as PR comment
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          path: ts/shielder-sdk/coverage/coverage-report.md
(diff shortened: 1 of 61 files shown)


.github/workflows/manual-publish-npm.yml
```diff
@@ -0,0 +1,99 @@
...
+        options:
+          - shielder-sdk
+          - shielder-sdk-crypto
+          - shielder-sdk-crypto-wasm
+      tag:
+        description: "NPM tag"
+        required: true
+        type: choice
+        options:
+          - develop
```
guspiel, line 20: just making sure that `develop` is an actual tag and not a misspelling of `development`

.github/workflows/manual-publish-npm.yml
```diff
@@ -0,0 +1,99 @@
...
+        working-directory: ts
+
+      - name: Build package
+        run: |
+          if [ "${{ github.event.inputs.package }}" = "shielder-sdk" ]; then
+            pnpm build-package:sdk
+          elif [ "${{ github.event.inputs.package }}" = "shielder-sdk-crypto" ]; then
+            pnpm build-package:crypto
+          else
+            pnpm build-package:crypto-wasm
```
guspiel, line 88: nit: you could handle `crypto-wasm` in another `elif` and stop the workflow in the `else` branch, just to have slightly more bug-resistant code in case this script evolves

.github/workflows/manual-publish-dev-env.yml
```diff
@@ -1,134 +0,0 @@
----
-name: "Build and deploy development environment"
-
-on:
-  workflow_dispatch:
-
-jobs:
-  build-and-push-relayer:
-    name: Build and push shielder-relayer
```
guspiel, line 9: just making sure that the removal of a script to push relayer is deliberate

guspiel (approved):

.github/workflows/manual-publish-npm.yml
kroist: an actual tag

.github/workflows/manual-publish-dev-env.yml
kroist: yes, we don't need to push relayer this way. it was supposed only for image push to docker registry so devs can download it, not for deployments.

we now use staging envs, so we don't need this functionality

.github/workflows/manual-publish-npm.yml
kroist: yes, I believe the best scenario is to have separate jobs declarations for each package. but now let's keep it simple

PR #96, opened 2025-01-19
ZK-1337: Create http server which serves bindings functions

kroist:


Cargo.toml
```diff
@@ -67,9 +71,12 @@ wasm-bindgen-rayon = { version = "1.2.1" }
...
 halo2_solidity_verifier = { path = "crates/halo2-verifier" }
+macros-core = { path = "crates/macros-core" }
+macros = { path = "crates/macros" }
 powers-of-tau = { path = "crates/powers-of-tau" }
 shielder-account = { path = "crates/shielder-account" }
 shielder-contract = { path = "crates/shielder-contract" }
 shielder-relayer = { path = "crates/shielder-relayer" }
 shielder-setup = { path = "crates/shielder-setup" }
 type-conversions = { path = "crates/type-conversions" }
+shielder_bindings = { path = "crates/shielder_bindings" }
```
pmikolajczyk41, line 82: let's keep them sorted (I will look for some lint to check this automatically)

crates/macros-core/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "macros-core"
```
pmikolajczyk41, line 2: do we really need yet another crate? why can't its content belong to `macros`?

crates/macros/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "macros"
+version = "0.1.0"
+edition = "2021"
```
pmikolajczyk41, line 4: we can inherit many attributes from the workspace

PR #97, opened 2025-01-22
ZK-382: Implement deposit for token shortlist

kroist:
Implement Deposit function for token shortlist feature:
- New contract `ShielderV_0_1_0`
- New abstract contract `TokenList`
- return `tokenIndex` in `DepositEvent` https://github.com/Cardinal-Cryptography/zkOS-monorepo/blob/33a4a93e7bc9a6c26571ca9ae4dabced50d86dab/contracts/ShielderV0_1_0.sol#L73
- new function `depositERC20` https://github.com/Cardinal-Cryptography/zkOS-monorepo/blob/33a4a93e7bc9a6c26571ca9ae4dabced50d86dab/contracts/ShielderV0_1_0.sol#L232-L241
- functions in main contract for appending token to list and removing the last in case of errors https://github.com/Cardinal-Cryptography/zkOS-monorepo/blob/33a4a93e7bc9a6c26571ca9ae4dabced50d86dab/contracts/ShielderV0_1_0.sol#L387-L396
- Modified verifier generator
- Some tests for upgrades


Makefile
```diff
@@ -69,9 +79,17 @@ generate-verifier-contracts:
 	cargo run --release --bin halo2_solidity_verifier_generator
 	$(MAKE) format-contracts
 
+
+.PHONY: generate-verifier-contracts-v0_1_0
+generate-verifier-contracts-v0_1_0: # Generate relation verifier contracts for v0_1_0 contract
+generate-verifier-contracts-v0_1_0:
+	cd crates/halo2-verifier
+	cargo run --release --bin halo2_solidity_verifier_generator_v0_1_0
```
fbielejec, line 87: why are we versioning the binary? Wouldn't it be enough to version the Cargo.toml?

Cargo.toml
```diff
@@ -53,6 +53,8 @@ sha3 = { version = "0.10" }
 shellexpand = { version = "3.1.0" }
 # https://github.com/Cardinal-Cryptography/zkOS-circuits/tree/7743b2f084d80685041c1b16718e6388226e9631
 shielder-circuits = { git = "ssh://git@github.com/Cardinal-Cryptography/zkOS-circuits", rev = "7743b2f" }
+# https://github.com/Cardinal-Cryptography/zkOS-circuits/tree/4f413a87dac5ceb7cf7581691e1b799c2914cdcc
+shielder-circuits-v0_1_0 = { package = "shielder-circuits", git = "ssh://git@github.com/Cardinal-Cryptography/zkOS-circuits", rev = "4f413a8" }
```
fbielejec, line 57: again is the version here neccessary? I have the commit hash to check it, we could use ta as more natural way to control versions.

contracts/ShielderV0_1_0.sol
```diff
@@ -0,0 +1,398 @@
...
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+/**
+ * @title ShielderV0_1_0
+ * @author CardinalCryptography
+ * @custom:oz-upgrades-from Shielder
+ * @custom:oz-upgrades-unsafe-allow external-library-linking
+ */
+// solhint-disable-next-line contract-name-camelcase
+contract ShielderV0_1_0 is
```
fbielejec, line 24: this could be simply called Shielder, version is already in the constant. Forge puts the artifacts in the file name dir anyway

crates/halo2-verifier/Cargo.toml
```diff
@@ -11,6 +11,10 @@ path = "src/lib/lib.rs"
 name = "halo2_solidity_verifier_generator"
 path = "src/generator.rs"
 
+[[bin]]
+name = "halo2_solidity_verifier_generator_v0_1_0"
```
fbielejec, line 15: isn't the "halo2_solidity_verifier" version enough?

scripts/ShielderV0_1_0.s.sol
```diff
@@ -0,0 +1,65 @@
...
+import { Script, console2 } from "forge-std/Script.sol";
+
+import { ERC1967Proxy } from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
+import { Shielder } from "../contracts/Shielder.sol";
+import { ShielderV0_1_0 } from "../contracts/ShielderV0_1_0.sol";
+import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+
+/* solhint-disable no-console */
+// solhint-disable-next-line contract-name-camelcase
+contract DeployShielderV0_1_0Script is Script {
```
fbielejec, line 14: This script deploys ShielderV0_1_0,  it's implicit, I don't need the unwieldy name  

Most likely, as the system evolves, we will keep adding new script with additional migrations, rather than changing existing ones. What I would suggest is to consecutively number them, in the order they need to be applied:

```
- 1_DeployShielder.s.sol
- 2_MigrateAnonymityRevoking.s.sol
- 3_AddErc20Support.s.sol
```

Names are an example off course.  The goal is that you can e.g. run migration 1-2 and get the state on Mainnet, or 1-3 and replicate the Testnet (as an example), but basically the scripts are always additive

contracts/ShielderV0_1_0.sol
```diff
@@ -0,0 +1,398 @@
...
+        publicInputs[5] = tokenIndex;
+
+        bool success = DepositVerifier.verifyProof(proof, publicInputs);
+
+        if (!success) revert DepositVerificationFailed();
+
+        uint256 newNoteIndex = _addNote(newNote);
+        _registerNullifier(oldNullifierHash);
+
+        emit DepositNative(
```
fbielejec, line 278: wrong event name

contracts/TokenList.sol
```diff
@@ -0,0 +1,68 @@
...
+    /*
+     * Add a token to the list
+     */
+    function _addTokenToList(address _token) internal {
+        TokenListStorage storage $ = _getTokenListStorage();
+        $.tokenAddressByIndex[$.tokensNumber] = _token;
+        $.tokensNumber++;
+    }
+
+    function _removeLastToken() internal {
```
fbielejec, line 59: what's the purpose of this fn?

contracts/ShielderV0_1_0.sol
```diff
@@ -0,0 +1,398 @@
...
+     * Add a token to the list
+     */
+    function addTokenToList(address _token) external onlyOwner {
+        _addTokenToList(_token);
+    }
+
+    /*
+     * Remove the last token from the list
+     */
+    function removeLastToken() external onlyOwner {
```
fbielejec, line 395: needed?

contracts/ShielderV0_1_0.sol
```diff
@@ -0,0 +1,398 @@
...
+            revert ContractBalanceLimitReached();
+        }
+        // @dev needs to match the order in the circuit
+        uint256[] memory publicInputs = new uint256[](5);
+        publicInputs[0] = idHiding;
+        publicInputs[1] = merkleRoot;
+        publicInputs[2] = oldNullifierHash;
+        publicInputs[3] = newNote;
+        publicInputs[4] = amount;
+        publicInputs[5] = tokenIndex;
```
fbielejec, line 269: a lot of this logic is shared with the native case, maybe refactor it out?

fbielejec (commented):
the implementation looks very good, bu the versioning is attacking from all angles :smile:  As much as I can understand naming the contracts with the semversion included, imo it's not needed all over the place, - it shoud be enough  to do semversioning in the tags and in the `Cargo.toml`, which has the version field for this exact purpose

contracts/ShielderV0_1_0.sol
```diff
@@ -0,0 +1,398 @@
...
+        uint256 index = _addNote(newNote);
+        _registerNullifier(idHash);
+
+        emit NewAccountNative(CONTRACT_VERSION, idHash, amount, newNote, index);
+    }
+
+    /*
+     * Make a native token deposit into the Shielder
+     */
+    function depositNative(
```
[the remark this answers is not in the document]
kroist: delegated to a shared helper

contracts/ShielderV0_1_0.sol
kroist: delegated to a shared helper

contracts/ShielderV0_1_0.sol
kroist: renamed to `Deposit`

Cargo.toml
kroist: we need to simultaneously have two versions of `shielder-circuits`:
- for bindings which have to support both old and new circuits
- and for deployment of a contract with old circuits verifiers

contracts/ShielderV0_1_0.sol
kroist: renamed

crates/halo2-verifier/Cargo.toml
kroist: binary targets are not versionable and must have unique names. There is a way of versioning it with feature, but it's hard to add `#[cfg]`'s into `generator.rs`, as lots of signatures from `shielder_circuits` have changed

Makefile
kroist: binary targets are not versionable and must have unique names. There is a way of versioning it with feature, but it's hard to add #[cfg]'s into generator.rs, as lots of signatures from shielder_circuits have changed

contracts/ShielderV0_1_0.sol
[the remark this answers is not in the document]
kroist: renamed to `Shielder` but kept in `ShielderV0_1_0.sol` file

crates/halo2-verifier/Cargo.toml
fbielejec: right - but what I don't understand is _**why**_ is the generator binary versioned, what is the purpose?

kroist: for having a possibility to generate both old and new verifiers. If you insist, I can move the functionality of generating both old and new bindings to a single generator binary

scripts/ShielderV0_1_0.s.sol
kroist: added `2_AddTokenSupport.sol` and script for deployment, first script is untouched at the moment

contracts/ShielderV0_1_0.sol
kroist: refactored to `setToken`

contracts/ShielderV0_1_0.sol
```diff
@@ -0,0 +1,398 @@
...
+            newNote,
+            newNoteIndex,
+            0
+        );
+    }
+
+    /*
+     * Make an ERC-20 token deposit into the Shielder
+     */
+    function depositERC20(
```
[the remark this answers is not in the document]
kroist: note sure it's currently meaningful to write integration tests without withdraw. let's leave it for another PR

fbielejec (approved):

contracts/ShielderV0_1_0.sol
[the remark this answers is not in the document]
guspiel: My point here is to change existing code while making sure that existing tests pass. Currently you're adding code outside the existing testing framework, is there any good reason for it?

contracts/ShielderV0_1_0.sol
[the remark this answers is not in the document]
guspiel: The new deposit is compatible with the old withdraw and new account, as long as you only use token index 0. The tests I have in mind do not use withdraw.

guspiel (changes requested):
Left comments noting that code is being added outside the existing testing framework; it seems preferable to iterate while keeping the tests.

contracts/ShielderV0_1_0.sol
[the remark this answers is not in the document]
kroist: rewriting tests for new circuits is out of scope of this PR. let's not make it giant.

PR #98, opened 2025-01-22
ZK-663: Bindings for new deposit circuit

kroist:
In this PR we create bindings for `DepositCircuit` from `shielder-circuits-v0_1_0`.

### DIFF
--- a/crates/shielder_bindings/build.rs
+++ b/crates/shielder_bindings/build.rs
@@ -52,7 +52,7 @@ fn gen_withdraw(full_params: &Params) {
     gen_params_pk::<WithdrawCircuit<F, RANGE_PROOF_CHUNK_SIZE>>("withdraw", full_params);
 }
 
-fn gen_depositv_0_1_0(full_params: &Params) {
+fn gen_deposit_v0_1_0(full_params: &Params) {
     gen_params_pk::<DepositCircuitV0_1_0>("deposit_v0_1_0", full_params);
 }
 
@@ -67,5 +67,5 @@ fn main() {
     gen_deposit(&full_params);
     gen_new_account(&full_params);
     gen_withdraw(&full_params);
-    gen_depositv_0_1_0(&full_params);
+    gen_deposit_v0_1_0(&full_params);
 }


crates/shielder_bindings/build.rs
```diff
@@ -42,24 +43,29 @@ fn gen_deposit(full_params: &Params) {
...
     gen_params_pk::<NewAccountCircuit<F>>("new_account", full_params);
 }
 
 /// This function is used to generate the artifacts for the WithdrawCircuit
-fn generate_withdraw(full_params: &Params) {
+fn gen_withdraw(full_params: &Params) {
     gen_params_pk::<WithdrawCircuit<F, RANGE_PROOF_CHUNK_SIZE>>("withdraw", full_params);
 }
 
+fn gen_depositv_0_1_0(full_params: &Params) {
```
pmikolajczyk41, line 55: nit: `gen_deposit_v0_1_0`

crates/shielder_bindings/src/utils.rs
```diff
@@ -8,6 +8,7 @@ use shielder_circuits::{
     poseidon::off_circuit::hash,
     FieldExt, F,
 };
+use shielder_circuits_v0_1_0::consts::NUM_TOKENS;
```
pmikolajczyk41, line 11: something is no yes - the 0.1.0 version is the one already deployed and it doesn't have NUM_TOKENS

kroist: `v0.1.0` is the version of a new contract 😛 , the deployed one is `v0.0.1`

It would be nice to have only one version of circuits dependency in the whole repo, but we would need to keep it at least for `shielder-sdk`, as it would need to have both circuits version during contract upgrade.

We can rename the dependency to `shielder_circuits_next`.

crates/shielder_bindings/build.rs
kroist: fixed

pmikolajczyk41 (approved):

PR #99, opened 2025-01-30
Relayer: permissible ERC20 tokens

pmikolajczyk41:
1. Add a configuration parameter to specify which ERC20 tokens are supported as payment tokens.
2. Add a parameter to the `RelayQuery` and validate it.

### DIFF
--- a/crates/shielder-relayer/src/config/cli.rs
+++ b/crates/shielder-relayer/src/config/cli.rs
@@ -144,7 +144,8 @@ pub struct CLIConfig {
     #[clap(
         long,
         help = "Addresses of the ERC20 tokens that are qualified as a fee token.",
-        long_help = "Addresses of the ERC20 tokens that are qualified as a fee token."
+        long_help = "Addresses of the ERC20 tokens that are qualified as a fee token. If not provided, the value from the \
+            environment variable `{FEE_TOKEN_ENV}` will be used. If that is not set, assumed to be empty."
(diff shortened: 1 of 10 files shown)


crates/shielder-relayer/src/config/cli.rs
```diff
@@ -140,4 +140,12 @@ pub struct CLIConfig {
...
     )]
     pub relay_gas: Option<u64>,
+
+    #[clap(
+        long,
+        help = "Addresses of the ERC20 tokens that are qualified as a fee token.",
+        long_help = "Addresses of the ERC20 tokens that are qualified as a fee token. If not provided, the value from the \
+            environment variable `{FEE_TOKEN_ENV}` will be used. If that is not set, assumed to be empty."
+    )]
+    pub fee_tokens: Option<Vec<String>>,
```
kroist, line 150: pls mention that the format is csv 🙏 we have so many envs that it's really hard to track their format

kroist (approved):

kroist:
To be aligned:

The relayer would generally accept any whitelisted token as a withdraw fee re-payment? And this re-payment correctness logic would be inside the smart contract raw call, correct?

crates/shielder-relayer/src/config/cli.rs
pmikolajczyk41: mentioned

PR #108, opened 2025-02-12
Bump circuits to cdd411

pmikolajczyk41:
- adapt for new `generate_keys_with_min_k` API
- add revoker pkey argument for new account:
  - hardcoded for integration tests
  - hardcoded for stress testing
  - additional command argument for CLI
- add token address to all shielder calls - everywhere hardcoded to a native token address (0); this is a temporary solution; once we are ready for supporting and testing different tokens, we will remove constant and introduce argument passing
- add mac salt to shielder account
- fix measurements workflow

### DIFF
--- a/contracts/AnonymityRevoker.sol
+++ b/contracts/AnonymityRevoker.sol
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.26;
+
+import { Initializable } from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+abstract contract AnonymityRevoker is Initializable {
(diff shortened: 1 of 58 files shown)


kroist (approved):

contracts/AnonymityRevoker.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.26;
+
+import { Initializable } from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+abstract contract AnonymityRevoker is Initializable {
+    // keccak256(abi.encode(uint256(keccak256("zkos.storage.AnonymityRevoker")) - 1)) & ~bytes32(uint256(0xff))
```
deuszx, line 7: What's the reason for this elaborate scheme?

PR #110, opened 2025-02-12
ZK-714: Split contract calls

guspiel:
Actually partially reusing some code @kroist wrote earlier.

Not using `trySafeTransfer` as @fbielejec suggested, because it is a very recent OpenZeppelin feature, not present in the newest 5.2.0 release.

Also making small fixes to the `gas-consumption` workflow: naming and making sure the binary is rebuilt in both branches, so that in can talk to the contract even if it changes.

I will add tests in subsequent PRs.

### DIFF
--- a/contracts/Shielder.sol
+++ b/contracts/Shielder.sol
@@ -12,6 +12,9 @@ import { Ownable2StepUpgradeable } from "@openzeppelin/contracts-upgradeable/acc
 import { PausableUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/PausableUpgradeable.sol";
 import { UUPSUpgradeable } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+using SafeERC20 for IERC20;
 
 /// @title Shielder
 /// @author CardinalCryptography
@@ -87,7 +90,6 @@ contract Shielder is
     error FeeHigherThanAmount();
     error MerkleRootDoesNotExist();
     error NativeTransferFailed();
-    error ERC20TransferFailed();
     error WithdrawVerificationFailed();
     error NewAccountVerificationFailed();
     error ZeroAmount();
@@ -95,7 +97,6 @@ contract Shielder is
     error ContractBalanceLimitReached();
     error WrongContractVersion(bytes3 actual, bytes3 expectedByCaller);
     error NotAFieldElement();
-    error IncorrectNativeAmount();
 
     modifier restrictContractVersion(bytes3 expectedByCaller) {
         if (expectedByCaller != CONTRACT_VERSION) {
@@ -135,10 +136,62 @@ contract Shielder is
     }
 
     /*
-     * Creates a fresh note, with an optional token deposit.
+     * Creates a fresh note, with an optional native token deposit.
      *
-     * This transaction serves as the entrypoint to the Shielder.
+     * This transaction serves as an entrypoint to the Shielder.
      */
+    function newAccountNative(
+        bytes3 expectedContractVersion,
+        uint256 newNote,
+        uint256 idHash,
+        bytes calldata proof
+    ) external payable whenNotPaused {
+        uint256 amount = msg.value;
+        // `address(this).balance` already includes `msg.value`.
+        if (address(this).balance > MAX_CONTRACT_BALANCE) {
+            revert ContractBalanceLimitReached();
+        }
+
+        newAccount(
+            expectedContractVersion,
+            address(0),
+            amount,
+            newNote,
+            idHash,
+            proof
+        );
+    }
+
+    /*
+     * Creates a fresh note, with an optional ERC20 token deposit.
+     *
+     * This transaction serves as an entrypoint to the Shielder.
+     */
+    function newAccountERC20(
+        bytes3 expectedContractVersion,
+        address tokenAddress,
+        uint256 amount,
+        uint256 newNote,
+        uint256 idHash,
+        bytes calldata proof
+    ) external whenNotPaused {
+        IERC20 token = IERC20(tokenAddress);
+        token.safeTransferFrom(msg.sender, address(this), amount);
+
+        if (token.balanceOf(address(this)) > MAX_CONTRACT_BALANCE) {
+            revert ContractBalanceLimitReached();
+        }
+
+        newAccount(
+            expectedContractVersion,
+            tokenAddress,
+            amount,
+            newNote,
+            idHash,
+            proof
+        );
+    }
+
     function newAccount(
         bytes3 expectedContractVersion,
         address tokenAddress,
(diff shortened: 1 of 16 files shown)


contracts/Shielder.sol
```diff
@@ -135,10 +136,62 @@ contract Shielder is
...
+    ) external payable whenNotPaused {
+        uint256 amount = msg.value;
+        // `address(this).balance` already includes `msg.value`.
+        if (address(this).balance > MAX_CONTRACT_BALANCE) {
+            revert ContractBalanceLimitReached();
+        }
+
+        newAccount(
+            expectedContractVersion,
+            address(0),
```
kroist, line 157: let's move it into a constant `NATIVE_TOKEN_NOTE_ADDRESS=address(0) `

contracts/Shielder.sol
```diff
@@ -147,15 +200,13 @@ contract Shielder is
         uint256 idHash,
         bytes calldata proof
     )
-        external
-        payable
+        internal
         whenNotPaused
```
kroist, line 204: you can remove this modifier

contracts/Shielder.sol
```diff
@@ -287,13 +459,6 @@ contract Shielder is
...
         _registerNullifier(oldNullifierHash);
 
-        handleTokenTransferOut(
-            tokenAddress,
-            withdrawalAddress,
-            amount - relayerFee
-        );
-        handleTokenTransferOut(tokenAddress, relayerAddress, relayerFee);
-
         emit Withdraw(
```
kroist, line 462: emit the event in the end of the function call -> after the token transfer is done

contracts/Shielder.sol
```diff
@@ -182,8 +233,62 @@ contract Shielder is
...
+        uint256 merkleRoot,
+        bytes calldata proof
+    ) external payable whenNotPaused {
+        uint256 amount = msg.value;
+        if (address(this).balance > MAX_CONTRACT_BALANCE) {
+            revert ContractBalanceLimitReached();
+        }
+        deposit(
+            expectedContractVersion,
+            address(0),
```
kroist, line 252: use constant

contracts/Shielder.sol
```diff
@@ -234,8 +336,79 @@ contract Shielder is
...
+        uint256 oldNullifierHash,
+        uint256 newNote,
+        bytes calldata proof,
+        address relayerAddress,
+        uint256 relayerFee
+    ) external whenNotPaused {
+        withdraw(
+            expectedContractVersion,
+            idHiding,
+            address(0),
```
kroist, line 356: use constant

kroist:
merge in the main branch, there would be some changes to be done

contracts/Shielder.sol
```diff
@@ -194,16 +299,13 @@ contract Shielder is
         uint256 merkleRoot,
         bytes calldata proof
     )
-        external
-        payable
+        internal
```
fbielejec, line 311: it's often customary to call private/internal with the `_` suffix, e.g. `_deposit`

contracts/Shielder.sol
```diff
@@ -147,15 +200,13 @@ contract Shielder is
         uint256 idHash,
         bytes calldata proof
     )
-        external
-        payable
+        internal
```
fbielejec, line 208: unless you plan on inheriting from this contract i suggest `private` visibility. Might even see some slightly lower gas consumption.

contracts/Shielder.sol
guspiel: done

contracts/Shielder.sol
guspiel: done

contracts/Shielder.sol
guspiel: done

contracts/Shielder.sol
```diff
@@ -191,8 +248,62 @@ contract Shielder is
...
+        uint256 oldNullifierHash,
+        uint256 newNote,
+        uint256 merkleRoot,
+        bytes calldata proof
+    ) external whenNotPaused {
+        IERC20 token = IERC20(tokenAddress);
+        token.safeTransferFrom(msg.sender, address(this), amount);
+        if (token.balanceOf(address(this)) > MAX_CONTRACT_BALANCE) {
+            revert ContractBalanceLimitReached();
+        }
```
fbielejec, line 294: you should adhere to the Checks-Effects-Interactions principle this WILL be pointed out during the audit, let's not give easy wins out

contracts/Shielder.sol
guspiel: done

kroist (approved):

contracts/Shielder.sol
guspiel: done

contracts/Shielder.sol
guspiel: done

contracts/Shielder.sol
guspiel: done, but had to write the balance check in `depositERC20` and `newAccountERC20` a bit differently, please check if makes sense to you

contracts/Shielder.sol
guspiel: done

contracts/Shielder.sol
```diff
@@ -177,23 +256,94 @@ contract Shielder is
...
+        uint256 merkleRoot,
+        bytes calldata proof
+    ) external whenNotPaused {
+        IERC20 token = IERC20(tokenAddress);
+        if (
+            amount > MAX_CONTRACT_BALANCE ||
+            token.balanceOf(address(this)) + amount > MAX_CONTRACT_BALANCE
+        ) {
+            revert ContractBalanceLimitReached();
+        }
```
fbielejec, line 321: hmm, shouldn't this check be performed _after_ the transfer?

contracts/Shielder.sol
```diff
@@ -139,11 +143,88 @@ contract Shielder is
...
+        uint256 symKeyEncryption,
+        bytes calldata proof
+    ) external whenNotPaused {
+        IERC20 token = IERC20(tokenAddress);
+        if (
+            amount > MAX_CONTRACT_BALANCE ||
+            token.balanceOf(address(this)) + amount > MAX_CONTRACT_BALANCE
+        ) {
+            revert ContractBalanceLimitReached();
+        }
```
fbielejec, line 203: same, this check should probably be happening after the transfer from sender

fbielejec (approved):

contracts/Shielder.sol
fbielejec: looks good, maybe it could use safeMath for addition just to be safe

contracts/Shielder.sol
```diff
@@ -139,11 +143,88 @@ contract Shielder is
...
+        uint256 amount,
+        uint256 newNote,
+        uint256 idHash,
+        uint256 symKeyEncryption,
+        bytes calldata proof
+    ) external whenNotPaused {
+        IERC20 token = IERC20(tokenAddress);
+        if (
+            amount > MAX_CONTRACT_BALANCE ||
+            token.balanceOf(address(this)) + amount > MAX_CONTRACT_BALANCE
```
fbielejec, line 200: safeMath just to be safe here?

contracts/Shielder.sol
```diff
@@ -177,23 +256,94 @@ contract Shielder is
...
+        uint256 idHiding,
+        uint256 oldNullifierHash,
+        uint256 newNote,
+        uint256 merkleRoot,
+        bytes calldata proof
+    ) external whenNotPaused {
+        IERC20 token = IERC20(tokenAddress);
+        if (
+            amount > MAX_CONTRACT_BALANCE ||
+            token.balanceOf(address(this)) + amount > MAX_CONTRACT_BALANCE
```
fbielejec, line 318: safeMath or it could overflow?

fbielejec (approved):
might need safeMath for the overflow checks, although Sol >=0.8 is supposed to add it on generated bytecode so probably safe to skip

guspiel:
> might need safeMath for the overflow checks, although Sol >=0.8 is supposed to add it on generated bytecode so probably safe to skip

Overflows will not happen also due to `||` operator short-circuiting, but I agree that having it more explicit would be better.

PR #111, opened 2025-02-13
Ts bump circuits 2

kroist:
integration of shielder-sdk with bindings&contract from #108


pmikolajczyk41 (approved):
awesome, thanks

guspiel (approved):
contract API will change again after https://github.com/Cardinal-Cryptography/zkOS-monorepo/pull/110

PR #114, opened 2025-02-14
Bump circuits to 0ddf37 (mac inputs)

pmikolajczyk41:


ts/shielder-sdk/__tests/actions/deposit.test.ts
```diff
@@ -14,6 +14,8 @@ import { DepositAction } from "../../src/actions/deposit";
 import { AccountState } from "../../src/state";
 import { IContract, VersionRejectedByContract } from "../../src/chain/contract";
 import { SendShielderTransaction } from "../../src/client";
+import { bigint } from "zod";
+import { call } from "viem/actions";
```
kroist, line 18: are these leftovers?

ts/shielder-sdk/src/state/manager.ts
```diff
@@ -111,6 +122,7 @@ const emptyAccountState = (
   return {
     /// Since the private key is an arbitrary 32byte number, this is a non-reversible mapping
     id,
+    macSalt: Scalar.fromBigint(0n),
```
kroist, line 125: the initial idea was to keep `AccountState` as small as it can be and JIT-derive the needed values. Because `macSalt` would be deterministically generated from `nonce&id` in the future, let's remove it from here.

ts/shielder-sdk/src/actions/deposit.ts
```diff
@@ -62,6 +64,10 @@ export class DepositAction extends NoteAction {
     const hId = await this.cryptoClient.hasher.poseidonHash([state.id]);
     const idHiding = await this.cryptoClient.hasher.poseidonHash([hId, nonce]);
 
+    const macSalt = state.macSalt;
+    // temporary placeholder for MAC computation, will be exposed through bindings in the future
+    const macCommitment = Scalar.fromBigint(hexToBigInt("0x4141414141"));
+
```
kroist, line 70: pls refer to comment for `state/types.ts`. Let's generate `macSalt` in `generateCalldata` function and pass it here as an argument.

kroist (changes requested):

ts/shielder-sdk/__tests/actions/deposit.test.ts
pmikolajczyk41: yup, removed

ts/shielder-sdk/src/state/manager.ts
pmikolajczyk41: removed

kroist (approved):

PR #115, opened 2025-02-14
SDK: use public inputs bindings

kroist:
introduce rust binding for public inputs generation instead of generating them with lower-level primitives


crates/shielder_bindings/src/circuits/withdraw.rs
```diff
@@ -54,6 +58,43 @@ impl WithdrawCircuit {
         )
     }
 
+    #[allow(clippy::too_many_arguments)]
+    pub fn pub_inputs(
+        &self,
```
deuszx, line 63: From the POV of Rust - you don't need `&self` here (it's not used). Do you need it for the Wasm generated code?

ts/shielder-sdk/__tests/helpers.ts
```diff
@@ -140,6 +153,17 @@ class MockedDepositCircuit implements DepositCircuit {
     return Promise.resolve(new Uint8Array());
   }
 
+  pubInputs(values: DepositAdvice): Promise<DepositPubInputs> {
+    return Promise.resolve({
```
deuszx, line 157: Is using `Promise.resolve` speeding things up?

crates/shielder_bindings/src/circuits/new_account.rs
```diff
@@ -43,6 +47,33 @@ impl NewAccountCircuit {
...
+            token_address: vec_to_f(token_address),
+            anonymity_revoker_public_key: vec_to_f(anonymity_revoker_public_key),
+        };
+
+        let concat_vec = knowledge
+            .serialize_public_input()
+            .iter()
+            .flat_map(|value| field_to_bytes(*value))
+            .collect();
+        concat_vec
```
pmikolajczyk41, line 74: I suggest you could introduce an auxiliary handler and reuse it in 3 places; pros:
 - you hide boilerplate
 - you don't need to introduce a variable just to name the procedure

ts/shielder-sdk-crypto-wasm/src/circuits/deposit.ts
```diff
@@ -44,6 +46,35 @@ export class DepositCircuit
...
+      values.id.bytes,
+      values.nonce.bytes,
+      values.nullifierOld.bytes,
+      values.trapdoorOld.bytes,
+      values.accountBalanceOld.bytes,
+      values.tokenAddress.bytes,
+      values.path,
+      values.value.bytes,
+      values.nullifierNew.bytes,
+      values.trapdoorNew.bytes
```
pmikolajczyk41, line 63: is there any way that we can ensure that the order is correct? a struct with named fields would be perfect, but I guess that both `uniffi::export` and `wasm_bindgen` can't handle that; do you think that some kind of builder patter would be feasible?

ts/shielder-sdk-crypto-wasm/src/circuits/deposit.ts
```diff
@@ -44,6 +46,35 @@ export class DepositCircuit
...
+    );
+    const pubInputs = splitUint8(pubInputsBytes, 6).map(
+      (bytes) => new Scalar(bytes)
+    );
```
pmikolajczyk41, line 67: just to be sure - this wouldn't have worked that nicely if the inputs weren't simple scalars, right?

ts/shielder-sdk-crypto-wasm/src/circuits/deposit.ts
```diff
@@ -44,6 +46,35 @@ export class DepositCircuit
     );
   }
 
+  async pubInputs(values: DepositAdvice): Promise<DepositPubInputs> {
```
pmikolajczyk41, line 49: do we have to be `async` here?

ts/shielder-sdk-crypto-wasm/src/circuits/newAccount.ts
```diff
@@ -39,6 +41,31 @@ export class NewAccountCircuit
...
+      (bytes) => new Scalar(bytes)
+    );
+
+    return Promise.resolve({
+      hNote: pubInputs[0],
+      hId: pubInputs[1],
+      initialDeposit: pubInputs[2],
+      tokenAddress: pubInputs[3],
+      anonymityRevokerPubkey: pubInputs[4],
+      symKeyEncryption: pubInputs[5]
```
pmikolajczyk41, line 66: an idea (not sure though how feasible) - what if `*PubInputs` have a method like `from_bytes_sequence` that would handle this split and guard ordering?

ts/shielder-sdk-crypto-wasm/src/circuits/withdraw.ts
```diff
@@ -45,6 +47,38 @@ export class WithdrawCircuit
...
+      values.tokenAddress.bytes,
+      values.path,
+      values.value.bytes,
+      values.nullifierNew.bytes,
+      values.trapdoorNew.bytes,
+      values.commitment.bytes
+    );
+    const pubInputs = splitUint8(pubInputsBytes, 7).map(
+      (bytes) => new Scalar(bytes)
+    );
```
pmikolajczyk41, line 69: I need guidance (I have some blackout) - `7` is passed as a `chunk_size`, but it seems to be actually number of chunks

ts/shielder-sdk-crypto/src/types.ts
```diff
@@ -42,14 +44,15 @@ export type DepositAdvice = {
   trapdoorNew: Scalar;
 };
 
+// follows the order in shielder-circuits::circuits::withdraw
 export type WithdrawPubInputs = {
```
pmikolajczyk41, line 48: maybe we could expose another, debug-only binding for ensuring the order?

ts/shielder-sdk/src/actions/deposit.ts
```diff
@@ -188,7 +157,7 @@ export class DepositAction extends NoteAction {
             scalarToBigint(pubInputs.idHiding),
             scalarToBigint(pubInputs.hNullifierOld),
             scalarToBigint(pubInputs.hNoteNew),
-            scalarToBigint(merkleRoot),
+            scalarToBigint(pubInputs.merkleRoot),
```
pmikolajczyk41, line 160: nice catch

pmikolajczyk41 (commented):
generally LGTM, I just have some minor questions/suggestions

crates/shielder_bindings/src/circuits/withdraw.rs
kroist: yep, moved it to pub function

ts/shielder-sdk/__tests/helpers.ts
kroist: no, it's to comply with an interface

crates/shielder_bindings/src/circuits/new_account.rs
kroist: introduced strucst for output values

ts/shielder-sdk-crypto-wasm/src/circuits/deposit.ts
kroist: introduced struct for output values

ts/shielder-sdk-crypto-wasm/src/circuits/deposit.ts
kroist: removed this as now using struct

ts/shielder-sdk-crypto-wasm/src/circuits/newAccount.ts
kroist: now using struct

ts/shielder-sdk-crypto/src/types.ts
kroist: now using struct, but the comment is still there (it might be helpful)

pmikolajczyk41 (approved):

PR #117, opened 2025-02-17
ZK-739: Price feed

obrok:


crates/powers-of-tau/Cargo.toml
```diff
@@ -16,4 +16,5 @@ byteorder = { workspace = true }
 halo2_proofs = { workspace = true }
 halo2curves = { workspace = true }
 num-bigint = { workspace = true }
+rand = { workspace = true }
```
pmikolajczyk41, line 19: after rebase, you can remove `dev-dependencies`

crates/shielder-relayer/src/price_feed.rs
```diff
@@ -0,0 +1,125 @@
...
+    Azero,
+    Btc,
+    Usdt,
+    Usdc,
+}
+
+/// A collection of prices for various coins.
+///
+/// The underlying structure is behind a mutex and a process to update it
+/// asynchronously can be strated with `start_price_feed`.
```
pmikolajczyk41, line 27:
```suggestion
/// asynchronously can be started with `start_price_feed`.
```

crates/shielder-relayer/src/price_feed.rs
```diff
@@ -0,0 +1,125 @@
...
+#[derive(Clone)]
+pub struct Prices {
+    validity: time::Duration,
+    refresh_interval: Duration,
```
pmikolajczyk41, line 31: why do we use two different `Duration`s? would it make sense to unify them?

crates/shielder-relayer/src/price_feed.rs
```diff
@@ -0,0 +1,125 @@
...
+        Self {
+            validity,
+            refresh_interval,
+            inner,
+        }
+    }
+
+    /// Get the price of a coin or `None` if the price is not available or outdated.
+    pub fn price(&self, coin: Coin) -> Option<Decimal> {
+        let inner = self.inner.lock().expect("Mutex poisoned");
```
pmikolajczyk41, line 71: is there an obvious reason to work with synchronous locking? in general, does it make sense to make `PriceFeed` async?

crates/shielder-relayer/src/price_feed.rs
```diff
@@ -0,0 +1,125 @@
...
+use serde::Deserialize;
+use strum::IntoEnumIterator as _;
+use strum_macros::EnumIter;
+use time::OffsetDateTime;
+use tokio::time::Duration;
+
+const BASE_PATH: &str = "https://api.diadata.org/v1/assetQuotation";
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, EnumIter)]
+pub enum Coin {
```
pmikolajczyk41, line 16: I suggest adding `Custom(&'static str)` variant, so that anyone can pass token path without changing relayer's code

pmikolajczyk41 (commented):
code looks fine, just few questions

crates/shielder-relayer/src/price_feed.rs
obrok: Hm, I just default to this. What is the rationale to prefer async? I took care so that the lock was taken out for just the time needed to get/update the map, so it shuoldn't ever block any tokio worker.

crates/shielder-relayer/src/price_feed.rs
pmikolajczyk41: I would say that we need to be flexible for adding new tokens easily - not sure though whether we should prefer changing code or adding runtime possibility

crates/shielder-relayer/src/price_feed.rs
pmikolajczyk41: to prefer async - everything else in relayer is async

crates/shielder-relayer/src/price_feed.rs
obrok: I imagine you'd want to set `refresh_interval` lower than `validity` (for example 1 min and 10 mins) so that:

1. There is no window where it's invalid but we have to wait to refresh
2. We get retries kind of "for free" - if an update fails, we just retry after the next `refresh_interval` and might still make it before `validity` elapses

crates/shielder-relayer/src/price_feed.rs
obrok: I don't trust the async mutexes, but sure, can change.

crates/shielder-relayer/src/price_feed.rs
obrok: It seems invalid to me to not have all of the config for a token in place - like missing the contract address or missing the price feed address and so on - so I was thinking of pushing the checks as close to compile-time as possible. WDYT?

crates/shielder-relayer/src/price_feed.rs
pmikolajczyk41: I meant - tokio's Duration and std's :)

crates/shielder-relayer/src/price_feed.rs
pmikolajczyk41: Fair enough

crates/shielder-relayer/src/price_feed.rs
obrok: Ah, the API exposes the same `Duration`. One is more convenient internally, so I do the conversion in the constructor instead of on every usage. The conversion is just a long, painful line, so I prefer to do it this way.

obrok:
```
        let validity =
            time::Duration::new(validity.as_secs() as i64, validity.subsec_nanos() as i32);
```

yuck

PR #126, opened 2025-02-21
ZK-776: Generate id on curve point

kroist:


fbielejec (approved):

crates/shielder-account/src/lib.rs
```diff
@@ -52,7 +52,7 @@ impl ShielderAccount {
     /// writing single-actor tests.
     pub fn new(id: U256) -> Self {
```
pmikolajczyk41, line 53: maybe we should rename this parameter to `id_seed`, so that it doesn't cause any confusion, that the `id` passed here might differ from the resulting;

also, the docstring should be updated

pmikolajczyk41 (approved):

crates/shielder-account/src/lib.rs
kroist: renamed

crates/shielder_bindings/src/secrets.rs
```diff
@@ -34,8 +35,11 @@ pub fn get_action_secrets(id: Vec<u8>, nonce: u32) -> ShielderActionSecrets {
 #[cfg_attr(feature = "build-wasm", wasm_bindgen)]
 #[cfg_attr(feature = "build-uniffi", uniffi::export)]
 pub fn derive_id(private_key_hex: &str, token_address_hex: &str) -> Vec<u8> {
-    u256_to_bytes(secrets::derive_id(
+    let id_u256 = secrets::derive_id(
```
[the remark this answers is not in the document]
kroist: renamed

PR #132, opened 2025-02-27
Generic types in `shielder-sdk-crypto`

kroist:
In this PR we:
- make types (such as `DepositAdvice`, `DepositPubInto`) in shielder-sdk-crypto generic, allowing them to be independent of `Scalar` object. This will be useful in mobile bindings.
- remove `CurvePointAffine`, as this wrapper doesn't bring much value. All integrated interfaces: wasm bindings, smart contracts and mobile bindings in future will have flat inputs, so structuring curve point just adds more wrapping/unwrapping logic.

### DIFF
--- a/ts/shielder-sdk/__tests/actions/deposit.test.ts
+++ b/ts/shielder-sdk/__tests/actions/deposit.test.ts
@@ -1,6 +1,5 @@
 import { beforeEach, describe, expect, vitest, it } from "vitest";
 import {
-  CryptoClient,
   DepositAdvice,
   DepositPubInputs,
   Scalar,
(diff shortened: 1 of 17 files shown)


ts/shielder-sdk/__tests/actions/deposit.test.ts
```diff
@@ -16,58 +15,6 @@ import { IContract, VersionRejectedByContract } from "../../src/chain/contract";
 import { SendShielderTransaction } from "../../src/client";
 import { nativeToken } from "../../src/types";
 
-const nativeTokenAddress = "0x0000000000000000000000000000000000000000";
-
-const expectPubInputsCorrect = async (
```
pmikolajczyk41, line 21: not used?

pmikolajczyk41 (approved):
I have a personal preference for keeping `CurvePointAffine`, but if it makes your work more onerous, then I'm okay with removing

ts/shielder-sdk/__tests/actions/deposit.test.ts
kroist: yep, not used

PR #138, opened 2025-02-28
CI: Measure verifier contract size change

pmikolajczyk41:


.github/workflows/_measure-gas-and-contract-size.yml
```diff
@@ -39,14 +42,29 @@ jobs:
...
 
       - name: Print gas measured on the current branch
-        run: cat current-report.txt
+        run: cat current-gas.txt
+
+      - name: Measure verifiers code size
+        run: |
+          echo -n "Deposit: " > current-size.txt
+          cat ./artifacts/DepositVerifier.sol/Halo2Verifier.json \
+            | jq -r '.bytecode.object' | wc -c >> current-size.txt
```
DamianStraszak, line 61: Maybe divide by 2 here to get size in bytes, because it's currently it's len of hex string.

DamianStraszak (changes requested):

DamianStraszak (approved):

.github/workflows/_measure-gas-and-contract-size.yml
pmikolajczyk41: done

PR #140, opened 2025-03-03
Extract proving keys&params from bindings

kroist:
- extracting halo2 proving key and params from bindings
- circuits now are initialized with `&[u8]` buffer, which translates to `Uint8Array` buffer in js env
- packing pk and params in separate entrypoint of `shielder-sdk-crypto-wasm`, example usage can be seen in e2e playwright tests:

https://github.com/Cardinal-Cryptography/zkOS-monorepo/blob/32da2b745782091025b6254f3b4fca0550f45955/ts/shielder-sdk-tests/web/EntryPoint.tsx#L21-L26

https://github.com/Cardinal-Cryptography/zkOS-monorepo/blob/32da2b745782091025b6254f3b4fca0550f45955/ts/shielder-sdk-tests/web/EntryPoint.tsx#L59-L85

### DIFF
--- a/crates/shielder_bindings/src/circuits/mod.rs
+++ b/crates/shielder_bindings/src/circuits/mod.rs
@@ -20,7 +20,7 @@ pub mod new_account;
 pub mod withdraw;
 
 pub trait WasmCircuit {
-    fn load_files() -> (Params, ProvingKey, u32);
+    fn load_files(params_buf: &[u8], pk_buf: &[u8]) -> (Params, ProvingKey, u32);
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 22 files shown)


ts/shielder-sdk-crypto-wasm/copy-keys.sh
```diff
@@ -0,0 +1,18 @@
...
+set -euo pipefail
+
+# Create dist-keys directory and subdirectories
+mkdir -p dist-keys/new_account
+mkdir -p dist-keys/deposit
+mkdir -p dist-keys/withdraw
+
+# Copy binary files from artifacts to dist-keys
+cp ../../crates/shielder_bindings/artifacts/new_account/params.bin dist-keys/new_account/
+cp ../../crates/shielder_bindings/artifacts/new_account/pk.bin dist-keys/new_account/
```
pmikolajczyk41, line 12:
```suggestion
cp ../../crates/shielder_bindings/artifacts/new_account/params.bin \
    ../../crates/shielder_bindings/artifacts/new_account/pk.bin \
    dist-keys/new_account/
```

pmikolajczyk41: just a suggestion

ts/shielder-sdk-crypto-wasm/src/wasmClientWorker.ts
```diff
@@ -38,12 +38,24 @@ expose(exposed);
...
  * @throws Will throw an error if the worker initialization fails.
  */
 export const initWasmWorker = async (
   threads: number,
+  new_account_params_buf: Uint8Array,
+  new_account_pk_buf: Uint8Array,
+  deposit_params_buf: Uint8Array,
+  deposit_pk_buf: Uint8Array,
+  withdraw_params_buf: Uint8Array,
+  withdraw_pk_buf: Uint8Array,
```
pmikolajczyk41, line 58: maybe we can wrap them into a struct?

ts/shielder-sdk-crypto-wasm/copy-keys.sh
kroist: changed

ts/shielder-sdk-crypto-wasm/src/wasmClientWorker.ts
kroist: wrapped

crates/shielder_bindings/src/circuits/mod.rs
```diff
@@ -20,7 +20,7 @@ pub mod new_account;
 pub mod withdraw;
 
 pub trait WasmCircuit {
-    fn load_files() -> (Params, ProvingKey, u32);
+    fn load_files(params_buf: &[u8], pk_buf: &[u8]) -> (Params, ProvingKey, u32);
```
DamianStraszak, line 23: Maybe rename `load_files` to `decode_from_bytes`

DamianStraszak (approved):

crates/shielder_bindings/src/circuits/mod.rs
kroist: renamed

PR #141, opened 2025-03-03
ZK-708: Make id depend on chain_id

kroist:
- add `chain_id` as a payload for `id` derivation.
- refactor `secrets::` module, breaks backward-compatibility of secrets, but we are allowed to do that before the release

### DIFF
--- a/crates/shielder-account/src/secrets.rs
+++ b/crates/shielder-account/src/secrets.rs
@@ -3,7 +3,7 @@ use rand::{rngs::OsRng, Rng};
 use sha3::Digest;
 use shielder_circuits::consts::NONCE_UPPER_LIMIT;
 
-pub enum Label {
+enum Label {
     Nullifier,
     Trapdoor,
     Id,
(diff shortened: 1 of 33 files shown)


crates/shielder-account/src/secrets.rs
```diff
@@ -3,9 +3,21 @@ use rand::{rngs::OsRng, Rng};
 use sha3::Digest;
 use shielder_circuits::consts::NONCE_UPPER_LIMIT;
 
-const NULLIFIER_LABEL: &[u8] = b"nullifier";
-const TRAPDOOR_LABEL: &[u8] = b"trapdoor";
-const ID_LABEL: &[u8] = b"id";
+pub enum Label {
```
pmikolajczyk41, line 6: if `pub`, then pls derive `Copy`, `Clone` and `Debug`

crates/shielder-account/src/secrets.rs
```diff
@@ -16,10 +28,9 @@ const FIELD_MODULUS: U256 = U256::from_limbs([
     0x30644e72e131a029,
 ]);
 
-fn hash(id: U256, label: &[u8], payload: &[u8]) -> U256 {
+fn hash(label: Label, payload: &[u8]) -> U256 {
     let mut hasher = sha3::Keccak256::new();
-    hasher.update(id.to_be_bytes_vec());
-    hasher.update(label);
+    hasher.update(label.as_bytes());
     hasher.update(payload);
```
pmikolajczyk41, line 34: to be precise, this is not compliant with our docs (`h(id, label, nonce)` vs `h(label, id, nonce)`)

crates/shielder-account/src/secrets.rs
```diff
@@ -28,22 +39,34 @@ fn hash(id: U256, label: &[u8], payload: &[u8]) -> U256 {
...
-pub fn derive_id(private_key: U256, token_address: Address) -> U256 {
-    hash(private_key, ID_LABEL, token_address.into_word().as_ref())
+pub fn derive_id(private_key: U256, chain_id: u64, token_address: Address) -> U256 {
+    // Concatenate chain_id bytes and token_address bytes
+    let mut payload = Vec::new();
+    payload.extend(private_key.to_be_bytes_vec());
+    payload.extend_from_slice(&chain_id.to_be_bytes());
+    payload.extend_from_slice(token_address.into_word().as_ref());
+
+    hash(Label::Id, &payload)
```
pmikolajczyk41, line 69: I get the idea with a common `hash` function that takes label and some arbitrary payload, but for the sake of explicit argument order, I'd give up this intermediate opaque payload

crates/shielder-account/src/secrets.rs
```diff
@@ -28,22 +39,34 @@ fn hash(id: U256, label: &[u8], payload: &[u8]) -> U256 {
...
 
 /// Private-key-dependent derivation of a per-token private ID.
-pub fn derive_id(private_key: U256, token_address: Address) -> U256 {
-    hash(private_key, ID_LABEL, token_address.into_word().as_ref())
+pub fn derive_id(private_key: U256, chain_id: u64, token_address: Address) -> U256 {
+    // Concatenate chain_id bytes and token_address bytes
+    let mut payload = Vec::new();
+    payload.extend(private_key.to_be_bytes_vec());
+    payload.extend_from_slice(&chain_id.to_be_bytes());
+    payload.extend_from_slice(token_address.into_word().as_ref());
```
pmikolajczyk41, line 67: according to the docs, we should get here account nonce instead of token address, shouldn't we?

pmikolajczyk41 (approved):

crates/shielder-account/src/secrets.rs
kroist: get your point, in that case I propose we get rid of `hash` function

crates/shielder-account/src/secrets.rs
kroist: made non-pub

crates/shielder-account/src/secrets.rs
kroist: reverted

crates/shielder-account/src/secrets.rs
kroist: look at it now. I think that previous `hash` function was really non-intuitive.

crates/shielder-account/src/secrets.rs
```diff
@@ -87,7 +109,6 @@ mod tests {
 
         let actual = derive_nullifier(U256::from(15), 0x000000ff);
 
-        assert_ne!(expected_before_modulo, actual);
```
pmikolajczyk41, line 90: are they equal now?

pmikolajczyk41 (approved):

crates/shielder-account/src/secrets.rs
kroist: they are, hash is less than modulo 🤷‍♂️

pmikolajczyk41: if I'm not mistaken, it should be fairly easy to find a new id, so that we go over modulo; this would make sense in our case (ensuring that the code indeed takes modulo)

kroist: added over-modulo check

crates/shielder-account/src/secrets.rs
```diff
@@ -3,9 +3,21 @@ use rand::{rngs::OsRng, Rng};
 use sha3::Digest;
 use shielder_circuits::consts::NONCE_UPPER_LIMIT;
 
-const NULLIFIER_LABEL: &[u8] = b"nullifier";
-const TRAPDOOR_LABEL: &[u8] = b"trapdoor";
-const ID_LABEL: &[u8] = b"id";
+enum Label {
```
guspiel, line 6: I consider constants better here for the sake of code simplicity, but feel free to leave it as it is.

guspiel (approved):
LGTM

PR #148, opened 2025-03-05
Sdk refactor 2

kroist:
- refactor `sync.ts` into smaller modules
- fix unit tests


guspiel (approved):
While I appreciate the excellent degree of code coverage, and I confirm that important edge cases of `StateTransitionFinder` are covered, I would suggest modifying the approach.

The tests focus less on actual inputs and outputs of the high-level account recovery process, and more on the interplay of small implementation details. We have tests of the form "class X makes steps A, B, C, D, E, F" with a lot of logic mocked away, and I don't see this as a guarantee that the account recovery produces a certain output given a certain chain state. 
 
1. These tests would significantly improve if we had access to Poseidon2 from TypeScript level -- which isn't difficult, we could just implement it. Mocking Poseidon2 makes us focus on very small pieces, and we don't verify if they fit together. It would be better to work with complete event lists with real hash values. Sure, we have excellent e2e tests, but they only address the positive code paths.
2. The volume of the tests is large, which costs a lot of reviewer time (and future readers'). I would look for ways to reduce code size, like addressing the fact that we have 1 class `StateTransitionFinder` with more significant logic and 2 small wrapper classes `HistoryFetcher` and `Synchronizer`, but we ended up with large test suites for each of them. A merge of these classes would help.
3. A drawback of this PR is that it focuses too much on implementation details. For example, the separation `StateTransitionFinder` -- `HistoryFetcher` -- `Synchronizer` makes test expectations focus on the interplay between them, like "StateSynchronizer called findStateTransition and then updateAccountState". My suggestion would be to merge these classes (and maybe some other as well) and try to build tests around a more high-level narrative like: "our account recovery given situation X returned Y with side effects A and B". 
4. Unit tests should focus on externally observable behavior and not implementation details. For example, we probably should not have tests verifying specifically that a mutex method was called. Leaving this uncovered is fine by me, but if we do want a test for that, it should probably take the form of some external caller calling the code 100x times simultaneously.

Please consider all comments in this PR optional.

PR #151, opened 2025-03-07
ZK-780: shielder-sdk: dynamically estimate gas for actions

kroist:


DamianStraszak (approved):

ts/shielder-sdk/src/chain/contract.ts
```diff
@@ -65,6 +64,11 @@ const getShielderContract = (
   });
 };
 
+type CalldataWithGas = {
```
guspiel, line 67: I know `view` calls it `gas`, but it would be helpful to the reader to call it "gas limit" on our side regardless, here and in other places.

guspiel (approved):
LGTM

ts/shielder-sdk/src/chain/contract.ts
guspiel: My point was that interaction with `viem` is just a small part of `shielder-sdk`, so keeping the name consistent is less important than the readability of our code.

But feel free to merge as it is.

PR #153, opened 2025-03-10
Configure native token in relayer

pmikolajczyk41:
Add a new configuration parameter - `native_token`, which specifies which coin is the native one.
In progress: Full documentation of the configuration, minor refactor of prices and coins.


crates/shielder-relayer/src/config/cli.rs
```diff
@@ -174,4 +174,12 @@ pub struct CLIConfig {
             the default value is `{DEFAULT_PRICE_FEED_VALIDITY_SECS:?}`."
     )]
     pub price_feed_validity: Option<u64>,
+
+    #[clap(
+        long,
+        help = "Token native to chain where the relayer operates.",
```
kroist, line 180: pls specify token format and a small list of tokens like "azero", "eth"

pmikolajczyk41: added a few examples - this shall be sufficient

PR #156, opened 2025-03-12
ZK-786: (draft) Add tests for withdrawERC20

guspiel:


crates/integration-tests/src/shielder/call_errors.rs
```diff
@@ -0,0 +1,91 @@
...
+    AmountOverDepositLimit(ShielderContract::AmountOverDepositLimit),
+    AmountTooHigh(ShielderContract::AmountTooHigh),
+    ContractBalanceLimitReached(ShielderContract::ContractBalanceLimitReached),
+    WrongContractVersion(ShielderContract::WrongContractVersion),
+    NotAFieldElement(ShielderContract::NotAFieldElement),
+
+    DestinationTriggeredRevert(),
+}
+
+impl From<ShielderContractErrors> for ShielderCallErrors {
```
fbielejec, line 25: looks like a place for the [thiserror](https://crates.io/crates/thiserror) crate?

fbielejec (approved):
looks good, non-blocking but could use the crate mentioned in one of the comments  to reduce amount of the hand-written code

crates/integration-tests/src/shielder/call_errors.rs
guspiel: agreed offline to leave it as it is for the sake of keeping enum structure flat

PR #157, opened 2025-03-12
Relayer: improve token config

pmikolajczyk41:
1. Rename `FeeToken` to `TokenKind` - this is an indicator, whether a coin is native or is deployed as an ERC20 contract (together with its address).
2. Rename and reorganize `TokenPricing*` to `TokenConfig` (structs, vars, envs). This holds full config for every token: coin id, kind, pricing.


kroist (approved):

kroist:
all good, but it would be really awkward to launch relayer with `NATIVE_TOKEN` different from the one in `TOKEN_CONFIG`

PR #160, opened 2025-03-13
Return account states list

kroist:
- add an endpoint which returns the list of all accounts state from storage, so the SDK user can easily retrieve them.
- remove `saveRawAccountAndIncrementNextAccountIndex` function and embed `accountIndex` increment in `saveRawAccount` function

### DIFF
--- a/ts/shielder-sdk/src/state/accountRegistry.ts
+++ b/ts/shielder-sdk/src/state/accountRegistry.ts
@@ -87,4 +87,18 @@ export class AccountRegistry {
 
     return null;
   }
+
+  async getAccountStatesList(): Promise<AccountStateMerkleIndexed[]> {
+    const accounts = await this.storageManager.getAllAccounts();
+
+    return Promise.all(
+      accounts.map(({ accountIndex, accountObject }) =>
+        this.accountStateSerde.toAccountState(
+          accountObject,
+          accountIndex,
+          getTokenByAddress(accountObject.tokenAddress as `0x${string}`)
+        )
+      )
+    );
+  }
 }
--- a/ts/shielder-sdk/src/state/accountRegistry.ts
+++ b/ts/shielder-sdk/src/state/accountRegistry.ts
@@ -59,7 +59,6 @@ export class AccountRegistry {
       accountState,
       accountIndex
     );
-    // Save
     await this.storageManager.saveRawAccount(accountIndex, accountObject);
   }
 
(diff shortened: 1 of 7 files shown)


ts/shielder-sdk/src/storage/storageManager.ts
```diff
@@ -20,6 +20,13 @@ export class StorageManager {
   async saveRawAccount(index: number, account: AccountObject): Promise<void> {
     const storageData = await this.storage.getStorage();
     storageData.accounts.set(index.toString(), { ...account });
+    if (index === storageData.nextAccountIndex) {
+      storageData.nextAccountIndex += 1;
+    } else if (index >= storageData.nextAccountIndex) {
+      throw new Error(
+        `Cannot save account at index ${index} when next account index is ${storageData.nextAccountIndex}`
```
guspiel, line 27: I think you mean `index > storageData.nextAccountIndex` and `Cannot save account at index ${index} when next account index is` **>>greater than<<** `${storageData.nextAccountIndex}`, right?

ts/shielder-sdk/src/storage/storageManager.ts
```diff
@@ -20,6 +20,13 @@ export class StorageManager {
   async saveRawAccount(index: number, account: AccountObject): Promise<void> {
     const storageData = await this.storage.getStorage();
     storageData.accounts.set(index.toString(), { ...account });
+    if (index === storageData.nextAccountIndex) {
+      storageData.nextAccountIndex += 1;
+    } else if (index >= storageData.nextAccountIndex) {
```
guspiel, line 25: nit: Consider putting the check before `storageData.accounts.set(index.toString(), { ...account });`. I know the `throw` prevents `storageData` from being saved, but it took me a minute to realize that.

ts/shielder-sdk/src/state/accountRegistry.ts
```diff
@@ -51,31 +51,16 @@ export class AccountRegistry {
...
-      await this.storageManager.saveRawAccountAndIncrementNextAccountIndex(
-        accountIndex,
-        accountObject
-      );
-    }
+    const accountObject = await this.accountStateSerde.toAccountObject(
+      accountState,
+      accountIndex
+    );
+    // Save
```
guspiel, line 62: nit: I'd really skip this comment now:P

guspiel (approved):
LGTM

ts/shielder-sdk/src/storage/storageManager.ts
kroist: good catch

ts/shielder-sdk/src/storage/storageManager.ts
kroist: moved the check to the top

ts/shielder-sdk/src/state/accountRegistry.ts
kroist: removed

PR #163, opened 2025-03-13
Relayer quote

pmikolajczyk41:
1. Quote endpoint expects arguments: pocket money and fee token.
2. `QuoteFeeResponse` was expanded to contain all information that might be useful for the client.
3. We compute fees according to the proposed document.
4. Changed /quote from GET to POST (to allow passing body)


crates/shielder-relayer/src/quote.rs
```diff
@@ -1,35 +1,120 @@
...
+        .token_config
+        .iter()
+        .find(|config| config.token_address().is_some_and(|a| a == address))
+        .ok_or("Fee token not found")?;
+    let fee_token_price = app_state
+        .prices
+        .price(fee_token.coin)
+        .ok_or("Fee token price not available")?;
+
+    let ratio = native_token_price / fee_token_price;
```
DamianStraszak, line 113: I'm wondering, the prices take decimals into account or not?

So, is price of ETH something like 2000*10^{-18} or rather 2000? 

Depending on the answer, this is either correct but weird, or incorrect :P

DamianStraszak (approved):

crates/shielder-relayer/src/quote.rs
pmikolajczyk41: discussed off-GH: I'll prepare a fix to the price feed

PR #167, opened 2025-03-17
Pocket money

pmikolajczyk41:
This PR introduces pocket money feature: during ERC20 withdrawal caller can pass some amount of native tokens (function is now payable). The value becomes part of the keccak commitment (to ensure that the proof creator intended for the same pocket money as the amount passed to the contract). Once proof is validated, whole amount will be forwarded to the withdrawal account.

Changes in this PR include:
 - contract logic
 - rust SDK adoption (including pocket money in the proof generation process)
 - ts SDK adoption (as above)

Things to notice:
 - when doing native withdrawal we put 0 to the keccak commitment (to stay consistent with ERC20 counterpart)
 - TS E2E tests use always 0 pocket money - this is because relayer still don't support this; I want to do it in a follow-up PR; this should be a minimal change where we can check and discuss basic flow and design


kroist (commented):
pocket money should be emitted in `Withdraw` event as well

kroist (commented):
shielder-sdk should include pocketMoney when syncing the withdraw transaction

ts/shielder-sdk-tests/web/fixtures/registrar.ts
```diff
@@ -32,13 +40,31 @@ export const setupRegistrar = (): RegistrarFixture => {
...
+    if (!tokenTxHistory.has("native")) {
+      tokenTxHistory.set("native", []);
+    }
+    tokenTxHistory.get("native")!.push({
+      type: "PocketMoney",
+      to,
+      amount: amount,
+      token: nativeToken()
+    });
+  };
```
kroist, line 56: where is this validated?

if we want to record the pocket money, it should be within the withdraw transaction scope.

ts/shielder-sdk-tests/tests/types.ts
```diff
@@ -89,9 +92,9 @@ export function recoverOp(): RecoverOp {
   };
 }
 
-// Shielder transaction representation, omitting the details like txHash, blockNumber, etc.
+// Shielder-related transaction representation, omitting the details like txHash, blockNumber, etc.
 export type ShortTx = {
-  type: "NewAccount" | "Deposit" | "Withdraw";
+  type: "NewAccount" | "Deposit" | "Withdraw" | "PocketMoney";
```
kroist, line 97: `PocketMoney` is not a separate transaction, but rather is in the scope of `Withdraw`

pmikolajczyk41: removed

ts/shielder-sdk-tests/web/fixtures/registrar.ts
pmikolajczyk41: removed

kroist:
All is left is to add pocketMoney to NoteEvent in contract/getNoteEventFromBlock

kroist (approved):

PR #168, opened 2025-03-18
Token config refined

pmikolajczyk41:


crates/shielder-relayer/src/token.rs
```diff
@@ -0,0 +1,23 @@
...
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+pub enum PriceProvider {
+    Url(String),
+    Static(Decimal),
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+pub struct Token {
+    pub kind: TokenKind,
+    pub decimals: u32,
```
kroist, line 21: This is ok, but imho decimals should be a responsibility of `TokenKind`, not `Token`:

```suggestion

    pub enum TokenKind {
        #[default]
        Native,
        ERC20(Address, u32),
    }

    pub struct Token {
        pub kind: TokenKind,
        pub price_provider: PriceProvider,
    }

    impl Token {
        fn decimals(&self) -> u32 {
            match self.kind {
                TokenKind::Native => 18,
                // bonus: you can actually fetch&cache ERC20 decimals
                // from the chain here, omitting configuring them manually
                TokenKind::ERC20(_, decimals) => decimals
            }
        }
    }
```

pmikolajczyk41: 1. you are assuming that every chain has 18 decimals for native - from yesterday discussion I got it that we shouldn't assume anything
2. for other parts of code it is easier to work with just information native/erc20, without additional config; in particular, decimal is only used for pricing; when the client makes a relay query, they shouldn't be bothered with specifying decimals for USDT; also, it'd be another check on our side if it is comatilbe with relayer config
3.

crates/shielder-relayer/src/quote.rs
```diff
@@ -95,21 +94,16 @@ struct Prices {
...
-        .iter()
-        .find(|config| config.token_address().is_some_and(|a| a == address))
-        .ok_or("Fee token not found")?;
     let fee_token_price = app_state
         .prices
-        .price(fee_token.coin)
+        .price(token)
         .ok_or("Fee token price not available")?;
 
     let ratio = native_token_price.unit_price / fee_token_price.unit_price;
```
kroist, line 109: this logic should be merged with the logic of `check_erc20_fee()` ratio finding logic

crates/shielder-relayer/src/relay/mod.rs
```diff
@@ -145,70 +142,45 @@ fn check_fee(
...
-            if add_fee_error_margin(expected_fee) < app_state.total_fee {
-                request_trace.record_insufficient_fee(query.fee_amount);
-                return Err(bad_request("Insufficient fee."));
-            }
-            Ok(())
-        }
+    let fee_token_price = get_price(fee_token)?.unit_price;
+    let native_price = get_price(TokenKind::Native)?.unit_price;
+    let ratio = fee_token_price / native_price;
+
```
kroist, line 168: this logic should be merged with the logic of `quote::get_token_price()` ratio finding logic

crates/shielder-relayer/src/quote.rs
pmikolajczyk41: this will be done in another PR; quoting is now isolated from relaying (we have just computation here, relay still operates on a fixed `total_fee`)

crates/shielder-relayer/src/relay/mod.rs
pmikolajczyk41: as above - this is now isolated intentionally

crates/shielder-relayer/src/price_feed/mod.rs
```diff
@@ -22,64 +21,71 @@ mod fetching;
...
             refresh_interval,
-            inner,
+            inner: Arc::new(Mutex::new(inner)),
         }
     }
 
-    /// Get the price of a coin or `None` if the price is not available or outdated.
-    pub fn price(&self, coin: Coin) -> Option<Price> {
+    /// Get the price of a token or `None` if the price is not available or outdated.
+    pub fn price(&self, token: TokenKind) -> Option<Price> {
```
kroist, line 56: what if there would be a race condition like:

- token is outdated
- quote fetches price, error is thrown
- price updater is executed

crates/shielder-relayer/src/price_feed/mod.rs
```diff
@@ -22,64 +21,71 @@ mod fetching;
...
+    pub fn new(tokens: &[Token], validity: Duration, refresh_interval: Duration) -> Self {
         let validity =
             time::Duration::new(validity.as_secs() as i64, validity.subsec_nanos() as i32);
-        let inner = Default::default();
+        let inner = tokens
+            .iter()
+            .map(|token| {
+                let price = match token.price_provider {
+                    PriceProvider::Url(_) => None,
+                    PriceProvider::Static(price) => Some(Price::new_now(price, token.decimals)),
```
kroist, line 42: why `None` for `Url` and `Some` for `Static` ?

crates/shielder-relayer/src/price_feed/mod.rs
```diff
@@ -22,64 +21,71 @@ mod fetching;
 pub struct Prices {
     validity: time::Duration,
     refresh_interval: Duration,
-    inner: Arc<Mutex<HashMap<Coin, Price>>>,
+    #[allow(clippy::type_complexity)]
+    inner: Arc<Mutex<HashMap<TokenKind, (Token, Option<Price>)>>>,
```
kroist, line 25: mapping between `TokenKind` and `Token` is static and doesn't change, yes? maybe it would make sense to have mutexed token->price and create a separate non-mutexed mapping for price feed lookup.

that would make the code much more readable by separating the concerns in the data structure.

crates/shielder-relayer/src/price_feed/fetching.rs
```diff
@@ -53,35 +50,76 @@ pub enum PriceFetchError {
...
 
-pub async fn fetch_price(coin: Coin) -> Result<Price, PriceFetchError> {
-    Ok(reqwest::get(format!("{BASE_PATH}{}", price_path(coin)))
-        .await?
-        .json::<PriceInfoFromProvider>()
-        .await
-        .map(|price_info| (coin, price_info).into())?)
+pub async fn fetch_price(token: &Token) -> Result<Price, PriceFetchError> {
+    let price_info = match &token.price_provider {
+        PriceProvider::Url(url) => {
```
kroist, line 55: for now it's good enough, but there is a high chance that in the future we would need to write lower-level wrappers for price providers.

crates/shielder-relayer/src/price_feed/mod.rs
pmikolajczyk41: then client gets info that price is temporarily unavailable and it should try again

crates/shielder-relayer/src/token.rs
kroist: I don't believe that we need native token to be configured, that's overkill. 

As for ERC-20, we actually don't need to configure anything but `token_address` and `price_feed`, as all other information is retrieve-able from chain (client's don't need to specify anything).

Imho the less you configure the better, but feel free to leave it as it is.

crates/shielder-relayer/src/price_feed/mod.rs
pmikolajczyk41: `Url` -> initially we don't know the price, because no fetching was initiated -> `None`
`Static` -> we already know the price -> `Some`

crates/shielder-relayer/src/price_feed/mod.rs
kroist: that's error-like behaviour, client should always get some price if the service is able to fetch it.

kroist:
actually, instead of mutexing the whole `inner` map, we can lazy-update the prices when needed.

pseudocode:

```rust
inner: Map<token, Mutex<Price>>;

let price = inner.get(&token);
if (is_valid(&price)) {
    return price;
}
else {
    let new_price = update_price(&token.provider);
    inner.update(&token, new_price);
    return new_price;
}
```

crates/shielder-relayer/src/price_feed/mod.rs
pmikolajczyk41: by what you describe, service is not able to provide a price; it doesn't have either current price or any guarantee that it can use the old price (because it is not valid anymore)

kroist: I'm saying that we need to do a force price fetch to prevent such scenario

pmikolajczyk41: we can _try_ fetching on demand, but there is no guarantee that the situation will resolve; I can add 1 trial in a separate PR for that

crates/shielder-relayer/src/token.rs
pmikolajczyk41: done as you wish

crates/shielder-relayer/src/price_feed/mod.rs
pmikolajczyk41: extracted

PR #173, opened 2025-03-19
AR POC

fbielejec:


crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -0,0 +1,216 @@
...
+pub async fn run(
+    rpc_url: &str,
+    shielder_address: &Address,
+    private_key_file: &str,
+    connection: Connection,
+) -> Result<(), CollectKeysError> {
+    let provider = create_simple_provider(rpc_url).await?;
+    let last_finalized_block_number = provider.get_block_number().await?;
+
+    let private_key = grumpkin::Fr::from_bytes(&private_key_bytes(private_key_file)?)
```
kroist, line 63: does not work with the key generated with big-endian option

crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -0,0 +1,216 @@
...
+        .into_option()
+        .ok_or(CollectKeysError::NotAGrumpkinBaseFieldElement)?;
+
+    db::create_viewing_keys_table(&connection)?;
+
+    for block_number in 0..=last_finalized_block_number {
```
kroist, line 69: block range should be configurable

crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -0,0 +1,216 @@
...
+        if let Some(block) = provider
+            .get_block_by_number(
+                BlockNumberOrTag::Number(block_number),
+                BlockTransactionsKind::Full,
+            )
+            .await?
```
kroist, line 75: this throws parsing errors on some L2 chains, like Arbitrum, Base, and even L1 Sonic Chain. Maybe we need to parse this manually or look for some advanced parsing in alloy.

crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -0,0 +1,216 @@
...
+    c2y: U256,
+    private_key: grumpkin::Fr,
+) -> Result<(), CollectKeysError> {
+    let ciphertext1 = GrumpkinPointAffine::new(u256_to_field(c1x), u256_to_field(c1y));
+    let ciphertext2 = GrumpkinPointAffine::new(u256_to_field(c2x), u256_to_field(c2y));
+
+    let GrumpkinPointAffine { x, .. } =
+        shielder_circuits::decrypt(ciphertext1.into(), ciphertext2.into(), private_key).into();
+
+    info!("Viewing key decoding {x:?}");
```
kroist, line 143: if it's sensitive information, don't `info!` it

crates/ar-cli/src/index_events.rs
```diff
@@ -0,0 +1,154 @@
...
+    shielder_address: &Address,
+    connection: Connection,
+) -> Result<(), IndexEventsError> {
+    let provider = create_simple_provider(rpc_url).await?;
+    let current_height = provider.get_block_number().await?;
+    let base_filter = Filter::new().address(*shielder_address);
+
+    db::create_events_table(&connection)?;
+
+    for block_number in (0..=current_height).step_by(BATCH_SIZE) {
```
kroist, line 55: block range should be configurable

crates/ar-cli/src/index_events.rs
```diff
@@ -0,0 +1,154 @@
...
+    providers::create_simple_provider,
+    ShielderContract::{Deposit, NewAccount, ShielderContractEvents, Withdraw},
+    ShielderContractError,
+};
+use thiserror::Error;
+use type_conversions::u256_to_field;
+
+use crate::db::{self, Event};
+
+const BATCH_SIZE: usize = 10_000;
```
kroist, line 22: batch size should be configurable, different RPCs have different limits

crates/ar-cli/src/revoke.rs
```diff
@@ -0,0 +1,80 @@
...
+        viewing_key,
+        tx_hash,
+        block_number,
+    } in events
+    {
+        if viewing_key.is_none() {
+            let commitment = blob_to_field(&mac_commitment)?;
+
+            for ViewingKey { viewing_key: key } in &keys {
+                let maybe_commitment = hash(&[blob_to_field(&mac_salt)?, blob_to_field(key)?]);
```
kroist, line 41: better `expected_` than `maybe_`

crates/ar-cli/src/collect_viewing_keys.rs
fbielejec: IMHO between the fact that there's a private key in plaintext,  there is an unencrypted db with all the txs deanonymized etc, etc this seriously does not matter

crates/ar-cli/src/index_events.rs
fbielejec: done

crates/ar-cli/src/index_events.rs
fbielejec: done

crates/ar-cli/src/revoke.rs
fbielejec: done

crates/ar-cli/src/collect_viewing_keys.rs
fbielejec: done

crates/ar-cli/src/collect_viewing_keys.rs
fbielejec: added an argument

crates/ar-cli/src/cli.rs
```diff
@@ -18,43 +19,100 @@ pub enum Endianess {
...
-fn parse_hex_as_seed(input: &str) -> Result<[u8; 32], &'static str> {
-    let mut decoded = [0u8; 32];
+#[derive(Debug, Args)]
+pub struct Db {
+    #[arg(long, default_value = "./ar_db.db3")]
+    pub path: PathBuf,
+}
+
+#[derive(Debug, Args)]
+pub struct Common {
```
pmikolajczyk41, line 93: suggested name: `ChainConfig` or something around this, but feel free to ignore

crates/ar-cli/src/cli.rs
```diff
@@ -18,43 +19,100 @@ pub enum Endianess {
...
+    #[arg(long, default_value = "0")]
+    pub from_block: u64,
+}
+
+fn parse_32byte_array(input: &str) -> Result<[u8; 32], &'static str> {
     let sanitized_input = if let Some(stripped) = input.strip_prefix("0x") {
         stripped
     } else {
         input
     };
```
pmikolajczyk41, line 109: `unwrap_or`?

crates/ar-cli/src/cli.rs
```diff
@@ -18,43 +19,100 @@ pub enum Endianess {
...
+    /// Read newAccount on-chain transactions and collect viewing keys
+    CollectKeys {
+        #[clap(flatten)]
+        common: Common,
+
+        #[clap(flatten)]
+        db: Db,
+
+        #[arg(long, default_value = "./private_key.bin")]
+        private_key_file: String,
```
pmikolajczyk41, line 63: `PathBuf`?

crates/ar-cli/src/generate.rs
```diff
@@ -35,10 +39,17 @@ pub fn run(seed: &[u8; 32], dir: PathBuf, endianess: cli::Endianess) -> Result<(
 
     let (private_key, public_key) = generate_keys(&mut rng);
 
-    debug!("private key: : {private_key:?}");
+    debug!(
+        "private key: : {private_key:?} [{}]",
+        field_to_u256(private_key)
+    );
```
pmikolajczyk41, line 45: do we want to print this to console? (some security concerns)

crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -0,0 +1,231 @@
...
+use type_conversions::u256_to_field;
+
+use crate::{
+    cli::Endianess,
+    db::{self, ViewingKey},
+};
+
+#[derive(Debug, Error)]
+#[error(transparent)]
+#[non_exhaustive]
```
pmikolajczyk41, line 31: what does it do?

crates/ar-cli/src/collect_viewing_keys.rs
pmikolajczyk41: do you know why? what kind of data cannot be parsed?

crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -0,0 +1,235 @@
...
+    let private_key = grumpkin::Fr::from_bytes(&bytes)
+        .into_option()
+        .ok_or(CollectKeysError::NotAGrumpkinBaseFieldElement)?;
+
+    db::create_viewing_keys_table(&connection)?;
+    // TODO: checkpoint tx blocks
+
+    for block_number in from_block..=last_finalized_block_number {
+        if let Some(block) = provider
+            .get_block_by_number(
```
pmikolajczyk41, line 90: to avoid that nesting, I recommend `let-else`

crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -0,0 +1,235 @@
...
+                                symKeyEncryptionC1X,
+                                symKeyEncryptionC1Y,
+                                symKeyEncryptionC2X,
+                                symKeyEncryptionC2Y,
+                                private_key,
+                            )?;
+                        }
+                    }
+                }
+            }
```
pmikolajczyk41, line 139: if we can't get block, we might log a warning about it

crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -0,0 +1,235 @@
...
+    for block_number in from_block..=last_finalized_block_number {
+        if let Some(block) = provider
+            .get_block_by_number(
+                BlockNumberOrTag::Number(block_number),
+                BlockTransactionsKind::Full,
+            )
+            .await?
+        {
+            if let Some(txs) = block.transactions.as_transactions() {
+                for tx in txs {
```
pmikolajczyk41, line 97: here the same suggestions - we can log lack of transactions; also, we can use `block.transactions.as_transactions().unwrap_or_else(|| log!(no txs); vec![])`

crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use alloy_primitives::U256;
+    use shielder_circuits::{poseidon::off_circuit::hash, Fr};
+    use type_conversions::u256_to_field;
+
+    #[test]
+    fn playground() {
```
pmikolajczyk41, line 197: why do we need this test?

crates/ar-cli/src/db.rs
```diff
@@ -0,0 +1,151 @@
...
+
+use rusqlite::{Connection, Result, Row};
+
+pub fn init(path: &PathBuf) -> Result<Connection, rusqlite::Error> {
+    Connection::open(path)
+}
+
+#[derive(Debug)]
+pub struct Event {
+    pub tx_hash: Vec<u8>,
```
pmikolajczyk41, line 12: `TxHash` from alloy? similarly `BlockNumber`

crates/ar-cli/src/db.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    let results = match viewing_key {
+        Some(key) => query.query_map([key], f)?,
+        None => query.query_map([], f)?,
+    };
+
+    let mut events = vec![];
+    for r in results {
+        events.push(r?);
+    }
```
pmikolajczyk41, line 105: afaik, vec of  result can be collected to result of vector

crates/ar-cli/src/db.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+pub fn create_events_table(connection: &Connection) -> Result<(), rusqlite::Error> {
+    connection.execute(
+        "CREATE TABLE IF NOT EXISTS events (
+            tx_hash BLOB PRIMARY KEY,
+            block_number INTEGER NOT NULL,
+            mac_salt BLOB NOT NULL,
+            mac_commitment BLOB NOT NULL,
+            viewing_key BLOB
```
pmikolajczyk41, line 43: 1. do we need/want to connect it to the other table?
2. this can be null, because initially, when we fetch an event, we don't know the viewing key, right?

crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -0,0 +1,235 @@
...
+    };
+
+    let private_key = grumpkin::Fr::from_bytes(&bytes)
+        .into_option()
+        .ok_or(CollectKeysError::NotAGrumpkinBaseFieldElement)?;
+
+    db::create_viewing_keys_table(&connection)?;
+    // TODO: checkpoint tx blocks
+
+    for block_number in from_block..=last_finalized_block_number {
```
pmikolajczyk41, line 88: would it make sense to emit encryption in events to avoid going through every single transaction in the history?

crates/ar-cli/src/db.rs
pmikolajczyk41: after full review, I can answer these questions myself, resolving

kroist (approved):

crates/ar-cli/src/collect_viewing_keys.rs
fbielejec: should be fixed now

crates/ar-cli/src/cli.rs
fbielejec: done

crates/ar-cli/src/cli.rs
fbielejec: done

crates/ar-cli/src/cli.rs
fbielejec: done

crates/ar-cli/src/generate.rs
fbielejec: added a redact

crates/ar-cli/src/collect_viewing_keys.rs
fbielejec: Errors may use error(transparent) to forward the source and Display methods straight through to an underlying error without adding an additional message

Basically means you get the underlying (error) message without the enclosing enum, which becomes opaque

crates/ar-cli/src/collect_viewing_keys.rs
fbielejec: we don;t removed

crates/ar-cli/src/collect_viewing_keys.rs
fbielejec: I don't know - needs a discussion, I;m working with what I have :)

crates/ar-cli/src/db.rs
fbielejec: good spot!

crates/ar-cli/src/db.rs
fbielejec: this is on purpose - I wanted a very basic struct here, that can be passed around

fbielejec:
> code lgtm, I'm about to play with it a bit
> 
> more general question - instead of collecting all keys, and revoking EVERYTHING, maybe it would be easier for us to change API to: "here is a `newAccount*` tx; please reveal this scoundrel"?

Technically this is the same amount of work - but i suppose we could have one `revoke-and-reveal` command? Depends which direction this tool goes to, there were ideas floated of using it for airdrop eligibility, so I'm hesitant on spending time on it not having a full picture

PR #174, opened 2025-03-20
Relayer: quote cache

pmikolajczyk41:


crates/shielder-relayer/src/quote_cache.rs
```diff
@@ -0,0 +1,227 @@
...
+                })
+                .or_insert(expiration);
+
+            cache.len()
+        };
+
+        // Once the storage reaches certain size, we try to run single garbage collection.
+        if cache_len >= CACHE_SIZE_THAT_TRIGGERS_GARBAGE_COLLECTION {
+            self.collect_garbage().await;
+        }
```
kroist, line 64: do we need this if we already have the garbage collection worker?

pmikolajczyk41: it's kinda two-dimensional protection: time-based triggered and size-based triggered; I thought it makes sense, but if you think it's an overkill, then I can remove it

deuszx: I'd be careful with checking the _total_ cache length rather than _per-token_. You don't want to garbage collect 	recent quotes b/c some other tokens' quotes triggered this.

pmikolajczyk41: I won't remove any valid quote; only expired

kroist: The goal is to prevent extreme cache growth. The problem that you are trying to solve with this approach is following:

> garbage collection worker fires every `quote_cache.validity * 10` milliseconds, which means that outdated requests from time range `[CURRENT_TIME - validity * 10, CURRENT_TIME)` are bloating the memory. There might be lots of such requests, and gc might be too slow for clearing them.

your strategy is that if the current number of queries exceeds the threshold, run GC more aggresively (at each insertion). 

I don't fully get a benefit of such hybrid approach and why we don't either:
- increase GC interval
- OR don't rely on GC at all, prune old queries when adding new queries

pmikolajczyk41: but I suppose this might be overengineering; unless we spot a problem with cache size, we shouldn't invest too much time for optimizing it now; maybe I will just leave the garbage collector worker - wdyt?

deuszx: Do you need more than one entry per `TokenKind`?

kroist: agree, let's keep it simple and predictable for now

kroist (approved):

PR #177, opened 2025-03-21
 Select nullifier block based on chain (special case for arbitrum chains)

kroist:


scripts/Shielder.s.sol
```diff
@@ -30,7 +32,13 @@ contract DeployShielderScript is Script {
...
         bytes memory data = abi.encodeCall(
             Shielder.initialize,
-            (owner, type(uint256).max, arPublicKeyX, arPublicKeyY) // set deposit limit to max
+            (
+                owner,
+                type(uint256).max,
+                arPublicKeyX,
+                arPublicKeyY,
+                isArbitrumChain
+            ) // set deposit limit to max
```
pmikolajczyk41, line 41: you can move the comment up

pmikolajczyk41 (approved):

PR #183, opened 2025-03-24
embed proving key&params on uniffi target

kroist:


fbielejec (approved):

crates/shielder_bindings/src/circuits/deposit.rs
```diff
@@ -54,15 +54,30 @@ impl From<DepositProverKnowledge<Fr>> for DepositPubInputsBytes {
...
     }
+}
 
+#[cfg(not(feature = "build-wasm"))]
+#[cfg_attr(feature = "build-uniffi", uniffi::export)]
+impl DepositCircuit {
+    #[cfg_attr(feature = "build-uniffi", uniffi::constructor)]
+    pub fn new_pronto() -> Self {
+        DepositCircuit(super::DepositCircuit::new_pronto(
+            include_bytes!("../../artifacts/deposit/params.bin"),
```
fbielejec, line 72: This blows up during compilation if the blobs are missing?

DamianStraszak (approved):

crates/shielder_bindings/src/circuits/deposit.rs
kroist: yes, but they always should be there because of `build.rs`

PR #186, opened 2025-03-26
Mobile bindings ci

kroist:


kroist:
1. these cjs files was a "bundled" yarn pkg manager, which was created by default when initalizing the repo. We don't need it, cause we can install it manually and have control over the version of yarn.
2. https://github.com/Cardinal-Cryptography/zkOS-monorepo/actions/runs/14081838438/job/39436385921 here is a successful action for the commit 872f0c09882d96c2b08bda65d9f5d555a7327884 . here I dry-run the packaging. In the next commit a9075c3ea604a5610912a5298b58e0cea6145bd0 I change the dry-run to the actual publishing.

pmikolajczyk41 (approved):

PR #188, opened 2025-03-26
Backoffs

fbielejec:


crates/ar-cli/src/main.rs
```diff
@@ -59,17 +67,44 @@ async fn main() -> Result<(), CliError> {
...
-                private_key_file,
-                endianess.clone(),
-                *from_block,
-                connection,
-                *redact_sensitive_data,
-            )
-            .await?
+            let mut delay = DEFAULT_BACKOFF;
+
+            loop {
```
pmikolajczyk41, line 72: I suggest introducing auxiliary function:
```rust
pub async fn retry_with_backoff<F, Fut, T>(mut op: F) -> Result<T, CollectKeysError>
where
    F: FnMut() -> Fut,
    Fut: std::future::Future<Output = Result<T, CollectKeysError>>,
{
    let mut delay = DEFAULT_BACKOFF;

    loop {
        match op().await {
            Ok(result) => return Ok(result),

            Err(CollectKeysError::Rpc(alloy_json_rpc::RpcError::Transport(
                TransportErrorKind::HttpError(http_err),
            ))) if http_err.is_rate_limit_err() => {
                delay = min(MAX_BACKOFF, delay * 2);
                info!("Rate limited. Waiting {:?} before retrying.", delay);
                sleep(delay);
            }

            Err(e) => return Err(e),
        }
    }
}
```

and calling it like:
```rust
retry_with_backoff(|| {
                collect_viewing_keys::run(
                    rpc_url,
                    shielder_address,
                    private_key_file,
                    endianess.clone(),
                    *from_block,
                    &db.path,
                    *redact_sensitive_data,
                )
            })
            .await?
```

here and below

pmikolajczyk41 (approved):

crates/ar-cli/src/main.rs
fbielejec: nice idea, done

PR #196, opened 2025-03-31
Ensure quotation has been queried before relay

pmikolajczyk41:
1. Remove margin tolerance for total fee
2. Ensure that the quote passed along the relay query is valid
3. Adapt CLI and SDK


crates/shielder-relayer/src/relay/mod.rs
```diff
@@ -174,7 +196,7 @@ fn check_erc20_fee(
 
     let expected_fee = scale_u256(fee_amount, ratio).map_err(server_error)?;
 
-    if add_fee_error_margin(expected_fee) < app_state.total_fee {
```
kroist, line 177: Why remove error margin?

kroist (approved):

PR #208, opened 2025-04-07
Relayer: check all endpoints in tests

pmikolajczyk41:
1. Use dedicated signer per testcase.
2. Split component tests into two modules
3. Unify endpoint return types (use JSON everywhere)
4. Add checks for all information that relayer provides through endpoints (like fee address, token config, API)


ts/shielder-sdk/src/chain/relayer.ts
```diff
@@ -204,6 +204,6 @@ export class Relayer implements IRelayer {
       throw new Error(`${responseText}`);
     }
 
-    return (await response.text()) as `0x${string}`;
+    return (await response.text()).replace(/"/g, "") as `0x${string}`;
```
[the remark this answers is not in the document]
pmikolajczyk41: done

ts/shielder-sdk/src/chain/relayer.ts
```diff
@@ -204,6 +204,6 @@ export class Relayer implements IRelayer {
       throw new Error(`${responseText}`);
     }
 
-    return (await response.text()) as `0x${string}`;
+    return (await response.json()) as `0x${string}`;
```
kroist, line 207: .address

CI will fail 🥲

kroist (approved):

ts/shielder-sdk/src/chain/relayer.ts
pmikolajczyk41: it doesn't; merging then

PR #215, opened 2025-04-15
Correctly parse relayer signer balances

pmikolajczyk41:
1. Do not use integer division. Do shifting on a raw string (to avoid any problems with `U256` -> `u128`/`u64` conversions)
2. Add a test for metrics format

### DIFF
--- a/crates/shielder-relayer/src/metrics.rs
+++ b/crates/shielder-relayer/src/metrics.rs
@@ -1,4 +1,4 @@
-use std::{ops::Div, time::Instant};
+use std::time::Instant;
 
 use anyhow::Result;
 use axum::{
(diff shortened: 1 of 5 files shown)


crates/shielder-relayer/src/metrics.rs
```diff
@@ -33,13 +32,14 @@ pub async fn prometheus_endpoint(
...
-            .parse::<f64>()
-            .unwrap_or(f64::NAN);
-        metrics::gauge!(SIGNER_BALANCES, "address" => signer.to_string()).set(tzero_balance);
+        let unit_balance = balance.read().await.unwrap_or_default().to_string();
+        let (tokens, decimals) = unit_balance.split_at(
+            unit_balance
+                .len()
+                .saturating_sub(NATIVE_TOKEN_DECIMALS as usize),
+        );
+        let token_balance = format!("{tokens}.{decimals}").parse::<f64>().unwrap();
```
DamianStraszak, line 41: String manipulation is quite error-prone. Instead you can just use this u256 -> f64 conversion function and then divide by 1e18. Nothing bad can happen to us, because precision loss is negligible.

pub fn u256_to_f64(value: U256) -> f64 {
	let pow264 = 2_f64.powi(64);
	value
		.into_limbs()
		.iter()
		.rev()
		.fold(0_f64, |acc, &limb| acc * pow264 + limb as f64)
}

pmikolajczyk41: applied, thanks!

PR #216, opened 2025-04-15
Improvements to AR-CLI

DamianStraszak:
Cleanes up the crate a little and adds possibility to use mnemonic phrases for priv key

### DIFF
--- a/crates/ar-cli/src/cli.rs
+++ b/crates/ar-cli/src/cli.rs
@@ -30,7 +30,6 @@ pub enum Command {
 
     InspectMnemonic {
         /// 12-word mnemonic phrase
-        #[arg(value_parser)]
         mnemonic: String,
     },
 
--- a/crates/ar-cli/src/cli.rs
+++ b/crates/ar-cli/src/cli.rs
@@ -12,7 +12,7 @@ pub struct Cli {
 
 #[derive(Debug, Subcommand)]
 pub enum Command {
-    /// Generate symmetric encryption keys
+    /// Generate encryption keys from a 32 byte seed
     Generate {
         /// Directory where the generated keys are to be written to.
         ///
@@ -27,18 +27,19 @@ pub enum Command {
         #[arg(long, value_parser = ValueParser::new(parse_32byte_array))]
         seed: [u8; 32],
     },
-
+    /// Inspect a 12-word mnemonic phrase
     InspectMnemonic {
         /// 12-word mnemonic phrase
         mnemonic: String,
     },
 
+    /// Inspect a 64-byte public key
     InspectPubkey {
         /// hex encoded 64-byte public key
         #[arg(value_parser = ValueParser::new(parse_64byte_array))]
         pk: [u8; 64],
     },
-
+    /// Generate a 12 word mnemonic and derive keys from it
     GenerateMnemonic,
 
     /// Read newAccount on-chain transactions and collect viewing keys
--- a/crates/ar-cli/src/cli.rs
+++ b/crates/ar-cli/src/cli.rs
@@ -10,13 +10,6 @@ pub struct Cli {
     pub command: Command,
 }
 
-#[derive(clap::ValueEnum, Clone, Default, Debug)]
-pub enum Endianess {
-    #[default]
-    LitteEndian,
-    BigEndian,
-}
-
 #[derive(Debug, Subcommand)]
 pub enum Command {
     /// Generate symmetric encryption keys
@@ -33,11 +26,21 @@ pub enum Command {
         /// Example: --seed 3fd54831f488a22b28398de0c567a3b064b937f54f81739ae9bd545967f3abab
         #[arg(long, value_parser = ValueParser::new(parse_32byte_array))]
         seed: [u8; 32],
-
-        /// Should the output be produced in the Lower Endian (the default) or Big Endian order?
-        #[clap(long, value_enum, default_value_t=Endianess::default())]
-        endianess: Endianess,
     },
+
+    InspectMnemonic {
+        /// 12-word mnemonic phrase
+        #[arg(value_parser)]
+        mnemonic: String,
+    },
+
+    InspectPubkey {
+        /// hex encoded 64-byte public key
+        #[arg(value_parser = ValueParser::new(parse_64byte_array))]
+        pk: [u8; 64],
+    },
+
+    GenerateMnemonic,
 
     /// Read newAccount on-chain transactions and collect viewing keys
     IndexEvents {
@@ -61,10 +64,6 @@ pub enum Command {
 
         #[arg(long, default_value = "./private_key.bin")]
         private_key_file: PathBuf,
-
-        /// is the key in the Lower Endian (the default) or Big Endian order?
-        #[clap(long, value_enum, default_value_t=Endianess::default())]
-        endianess: Endianess,
 
         #[arg(long, default_value = "true")]
         redact_sensitive_data: bool,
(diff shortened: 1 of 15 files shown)


crates/ar-cli/src/cli.rs
```diff
@@ -33,12 +26,22 @@ pub enum Command {
         /// Example: --seed 3fd54831f488a22b28398de0c567a3b064b937f54f81739ae9bd545967f3abab
         #[arg(long, value_parser = ValueParser::new(parse_32byte_array))]
         seed: [u8; 32],
+    },
+
+    InspectMnemonic {
+        /// 12-word mnemonic phrase
+        #[arg(value_parser)]
```
pmikolajczyk41, line 33: why do we use `value_parser` here?

crates/ar-cli/src/generate.rs
```diff
@@ -4,65 +4,52 @@ use std::{
...
-    let mut rng = ChaCha12Rng::from_seed(*seed);
-
-    info!("Generating key pair...");
+pub fn run(seed: &[u8; 32], dir: &PathBuf) -> Result<(), Error> {
+    run_inner(seed, Some(dir))
+}
 
-    let (private_key, public_key) = generate_keys(&mut rng);
 
+pub fn run_inner(seed: &[u8; 32], dir: Option<&PathBuf>) -> Result<(), Error> {
```
pmikolajczyk41, line 24: I guess it shouldn't be `pub`

crates/ar-cli/src/collect_viewing_keys.rs
```diff
@@ -48,18 +46,16 @@ pub async fn run(
...
-    let bytes = match endianess {
-        Endianess::LitteEndian => private_key_bytes(private_key_file)?,
-        Endianess::BigEndian => {
-            let mut bytes = private_key_bytes(private_key_file)?;
-            bytes.reverse();
-            bytes
-        }
+    let bytes = {
+        let mut bytes = private_key_bytes(private_key_file)?;
+        // We use big-endian encoding for the private key
```
pmikolajczyk41, line 51: I need help here - why?

crates/ar-cli/src/generate.rs
```diff
@@ -4,65 +4,52 @@ use std::{
...
-        field_to_u256(y)
-    );
-
-    let (private_key_bytes, x_bytes, y_bytes) = match endianess {
-        cli::Endianess::BigEndian => (private_key.to_bytes_be(), x.to_bytes_be(), y.to_bytes_be()),
-        cli::Endianess::LitteEndian => {
-            (private_key.to_bytes_le(), x.to_bytes_le(), y.to_bytes_le())
+    let public_key_bytes = serialize_pub_key(public_key);
+    let hex_pub_key = hex::encode(public_key_bytes);
+    println!("Public key: {}", hex_pub_key);
```
pmikolajczyk41, line 32: maybe we shouldn't mix `println` with `info`

crates/ar-cli/src/generate.rs
```diff
@@ -4,65 +4,52 @@ use std::{
...
-
-    info!("key pair files written to {dir:?}");
-
+        None => {
+        }
+    }
     Ok(())
 }
+
+pub fn run_mnemonic() -> Result<(), Error> {
```
pmikolajczyk41, line 47: having `run` and `run_mnemonic` as public API of the module might become a bit misleading if we have to go back do this code in a few months

crates/ar-cli/src/inspect.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use crate::{common::{deserialize_pub_key, mnemonic_to_seed, seed_to_keypair, serialize_pub_key}, error::Error};
+
+
+
+
+pub fn run_mnemonic(maybe_mnemonic: &str) -> Result<(), Error> {
+    let mnemonic = Mnemonic::from_phrase(maybe_mnemonic, Language::English)?;
+    let seed = mnemonic_to_seed(&mnemonic);
+    let (_private_key, public_key) = seed_to_keypair(&seed);
```
pmikolajczyk41, line 11: maybe we could print private key as well

scripts/deploy-shielder.sh
```diff
@@ -46,10 +46,17 @@ if [ -n "${AR_PUBLIC_KEY:-}" ]; then
...
     # Generate the public key using ar-cli
     echo "Generating AR key pair from seed..."
-    cargo run --bin ar-cli -- generate --seed ${AR_SEED} --endianess big-endian
+    cargo run --bin ar-cli -- generate --seed ${AR_SEED}
+    AR_PUBLIC_KEY=$(cat public_key.bin | cast from-bin)
+    AR_PUBLIC_KEY_X="0x$(echo $AR_PUBLIC_KEY | cut -c3-66)"
+    AR_PUBLIC_KEY_Y="0x$(echo $AR_PUBLIC_KEY | cut -c67-130)"
+    
+    # Assert correct length of AR_PUBLIC_KEY_X and AR_PUBLIC_KEY_Y
+    if [ ${#AR_PUBLIC_KEY_X} -ne 66 ] || [ ${#AR_PUBLIC_KEY_Y} -ne 66 ]; then
```
pmikolajczyk41, line 55: why 66?

crates/ar-cli/src/common.rs
```diff
@@ -0,0 +1,74 @@
...
+pub fn deserialize_pub_key(
+    bytes: &[u8],
+) -> Result<GrumpkinPoint<Fr>, Error> {
+    if bytes.len() != 64 {
+        return Err(Error::DeserializePubKey);
+    }
+    let x = blob_to_field(&bytes[0..32].iter().rev().copied().collect::<Vec<u8>>())?;
+    let y = blob_to_field(&bytes[32..64].iter().rev().copied().collect::<Vec<u8>>())?;
+    if y*y != x*x*x - Fr::from(17) {
+        return Err(Error::DeserializePubKey);
```
pmikolajczyk41, line 33: `PubkeyNotOnCurve` I guess

pmikolajczyk41 (approved):
minor comments, resolve at your will; generally lgtm

crates/ar-cli/src/collect_viewing_keys.rs
DamianStraszak: This is for consistency with pub key, and pub key is for easier integration with foundry, because EVM uses big-endian. This all doesn't matter, but as long as it's consistent we are good.

scripts/deploy-shielder.sh
DamianStraszak: 0x is there

DamianStraszak: And 64 hex symbols is 32 bytes, which represent a single U256 in big endian.

crates/ar-cli/src/inspect.rs
DamianStraszak: We could, but the intention here is to rely just on the mnemonic, we don't want to ask the AR to save the private key.

crates/ar-cli/src/generate.rs
DamianStraszak: I'm not sure what is a proper convention with the `run` function, but I changed the names to be more descriptive now, so hopefully it's better, and less chance for confusion.

crates/ar-cli/src/generate.rs
DamianStraszak: Yeah, removed one log here to stay with prints when generating the key.

crates/ar-cli/src/generate.rs
DamianStraszak: Correct, fixed.

crates/ar-cli/src/cli.rs
DamianStraszak: Honest reason: because I was copy-pasting shit, and missed it. Fixed :P

PR #230, opened 2025-05-13
Report price age and expiration

pmikolajczyk41:
1. Fix bug from the #228 - we shouldn't break the loop over all tokens if connecting to a price provider for one token failed. Now, we just `continue`
2. Expose `expired_price` metric - for all supported tokens, we indicate whether the price is outdated.
3. Expose `price_age` metric - for all supported tokens, display its price age in seconds. For static prices it is 0, for unavailable price it is `f64::MAX`.

Below is a screenshot for a relayer with 3 tokens:
 - native with static price
 - erc20 with real price feed url
 - erc20 with invalid feed url

image

### DIFF
--- a/crates/shielder-relayer/src/price_feed/mod.rs
+++ b/crates/shielder-relayer/src/price_feed/mod.rs
@@ -122,8 +122,8 @@ impl Prices {
 /// Start a price feed that updates the prices in the given `Prices` instance.
 pub async fn start_price_feed(prices: Prices) -> Result<(), anyhow::Error> {
     loop {
-        tokio::time::sleep(prices.refresh_interval).await;
         prices.update().await;
+        tokio::time::sleep(prices.refresh_interval).await;
     }
 }
 
--- a/crates/shielder-relayer/src/price_feed/mod.rs
+++ b/crates/shielder-relayer/src/price_feed/mod.rs
@@ -1,9 +1,7 @@
-use std::{
-    collections::HashMap,
-    sync::{Arc, Mutex},
-};
+use std::{collections::HashMap, sync::Arc};
 
 use fetching::fetch_price;
+use parking_lot::Mutex;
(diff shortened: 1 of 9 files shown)


crates/shielder-relayer/src/price_feed/mod.rs
```diff
@@ -57,6 +59,30 @@ impl Prices {
...
+    pub fn current_prices(&self) -> HashMap<TokenKind, Option<Price>> {
+        self.tokens.keys().map(|k| (*k, self.price(*k))).collect()
+    }
+
+    pub fn price_ages(&self) -> HashMap<TokenKind, Option<time::Duration>> {
+        let now = OffsetDateTime::now_utc();
+        self.inner
+            .iter()
+            .map(|(token, price)| {
+                let price = price.lock().expect("Mutex poisoned");
```
DamianStraszak, line 72: can use `parking_lot` crate Mutex, to get rid of unnecessary expects.

crates/shielder-relayer/src/price_feed/mod.rs
```diff
@@ -82,17 +116,13 @@ impl Prices {
                 .expect("Mutex poisoned")
                 .replace(price);
         }
-
-        Ok(())
     }
 }
 
 /// Start a price feed that updates the prices in the given `Prices` instance.
 pub async fn start_price_feed(prices: Prices) -> Result<(), anyhow::Error> {
```
DamianStraszak, line 123: can change the output type to `!`

DamianStraszak (approved):

crates/shielder-relayer/src/price_feed/mod.rs
pmikolajczyk41: unfortunately I can't - it is used within `tokio::try_join!`, which requires `Result`

PR #232, opened 2025-05-15
Fix Relayer workflows

pmikolajczyk41:
**Previously**
- 3 identical workflow files (difference only in a single line, specifying environment argument to an invoked action)
- `deploy` action touched old `shielder-relayer` configuration in aleph-apps (using Kustomize)
- possibility of changing Shielder contract address at the same time

**Now**
- single workflow with additional argument (dev / stage / prod / all)
- no option to change contract address (not needed & mixing two different responsibilities)
- simple `yq` substitution instead of kustomization (`shielder-relayer-v2` doesn't use overlays etc)


.github/actions/shielder-relayer/deploy/action.yml
```diff
@@ -21,15 +21,14 @@ inputs:
     description: Kustomize version
     required: true
   environment:
+    description: Environment to deploy to
     type: choice
     options:
+      - all (dev, stage, prod)
```
DamianStraszak, line 27: I see the intention behind the bracket, but can we make it just `all` -- I'm always scared to have whitespace in these kind of variables.

DamianStraszak (approved):
Looks good, but only running it will tell us if it works :P

.github/actions/shielder-relayer/deploy/action.yml
pmikolajczyk41: shortened

PR #238, opened 2025-05-20
Expose shielder account recovery

pmikolajczyk41:
1. expose method for computing `prenullifier` from account
2. expose utilities for fetching shielder actions from chain (recovery)
3. in CLI, allow passing optional seed for ZK ID


crates/shielder-cli/src/app_state.rs
```diff
@@ -82,12 +82,21 @@ impl AppState {
...
-            e.insert(ShielderAccount::new(field_to_u256(id_per_token), token));
+            if let Some(zkid_seed) = zkid_seed {
+                e.insert(ShielderAccount::new(zkid_seed, token));
+            } else {
+                let seed = U256::from_str(&self.signing_key)
+                    .expect("Invalid key format - cannot cast to U256");
+                let id_per_token = field_to_u256(hash(&[
+                    u256_to_field(seed),
+                    address_to_field(token.address()),
+                ]));
```
obrok, line 97: Suggestion: rewrite so that computing the default seed happens first, and then the actual `e.insert(...)` only appears once. Maybe even extract default seed computation to another function.

obrok (approved):

crates/shielder-cli/src/app_state.rs
pmikolajczyk41: suggestion applied, thanks

PR #244, opened 2025-06-05
add fee estimation for new_account & deposit

kroist:


crates/fee-estimator/README.md
```diff
@@ -0,0 +1,63 @@
...
+   cp .env.example .env
+   # Edit .env with your configuration
+   ```
+
+2. Required environment variables in `.env`:
+
+   - `RPC_URL` - RPC endpoint URL
+   - `CONTRACT_ADDRESS` - Shielder contract address (40-character hex string starting with 0x)
+   - `ERC20_TOKEN_ADDRESS` - ERC20 token address (40-character hex string starting with 0x)
+   - `ACCOUNT_PK` - Account private key, must have at least 1 gwei of native token and 2 smallest points of erc20-token at the first start.
```
JanKuczma, line 28: It also has to approve Shielder (`CONTRACT_ADDERSS`) as spender of those 2 smallest points of the erc20-token.

crates/fee-estimator/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+        interval.tick().await;
+
+        info!("Fetching new fees...");
+
+        // Update fees with mocked values that change each iteration
+        let new_fees = get_fee_values(&service_config).await;
+
+        if new_fees.is_err() {
+            error!("Failed to fetch new fees: {:?}", new_fees.err());
+            continue; // Skip this iteration if fetching fees fails
```
JanKuczma, line 48: Since this may happen repeatedly, maybe we should add `update_ts` to the `FeeResponse`, so the client of this service is aware how old the estimation is?

JanKuczma (commented):
This is looking very good. Just some small picks.

I think this service could handle a list of ERC20 tokens to make estimations, instead of just one. Unless there's no need for it (we assume ERC20 does not affect gas estimation) or some reason against.

crates/fee-estimator/src/main.rs
kroist: added

PR #253, opened 2025-06-27
Add support for lightweight (single-threaded) WASM in shielder-bindings

kroist:


crates/shielder_bindings/Makefile
```diff
@@ -40,6 +42,16 @@ remove-unused-files-wasm:
 	(cd pkg/pkg-web-singlethreaded && rm README.md package.json .gitignore)
 	(cd pkg/pkg-web-multithreaded && rm README.md package.json .gitignore)
 
+.PHONY: wasm-light
+wasm-light: # Build wasm bindings without circuits
+	rm -rf pkg-light
```
Marcin-Radecki, line 47: `clean-wasm` target should clear this directory (as it is a wildcard `rm -rf pkg*`), why do we need to clear artifacts here manually?

crates/shielder_bindings/Cargo.toml
```diff
@@ -43,6 +43,7 @@ multithreading-wasm = [
     "wasm-bindgen-rayon",
     "shielder-circuits/multithreading",
 ]
+circuits = []
```
Marcin-Radecki, line 46: I was wondering if it makes sense to put some comments here, like what feature means. E.g. comment for this one would be `# Build wasm bindings without circuits` etc

crates/shielder_bindings/src/lib.rs
```diff
@@ -8,6 +8,7 @@ pub use wasm_bindgen_rayon::init_thread_pool;
 
 extern crate alloc;
 
+#[cfg(feature = "circuits")]
```
Marcin-Radecki, line 11: Do  we have a target in CI that builds that feature? If not, I think we should, in general we should have in CI every matrix of the features combination we use in production with `cargo check`, just to make sure it builds on PR.

Marcin-Radecki, line 11: Can we add a section to this crate README, what the light WASM is about?

crates/shielder_bindings/Makefile
```diff
@@ -40,6 +42,16 @@ remove-unused-files-wasm:
...
+	wasm-pack build --release --target web --out-dir pkg-light/pkg-web-singlethreaded -- \
```
Marcin-Radecki, line 48: Why we ever create separate folder, why don't just have `--out-dir pkg/pkg-web-without-curcuits`? And more general qustion, why don't we put this build into existing `build-pkg-wasm`, and ship everything?

crates/shielder_bindings/Makefile
```diff
@@ -40,6 +42,16 @@ remove-unused-files-wasm:
 	(cd pkg/pkg-web-singlethreaded && rm README.md package.json .gitignore)
 	(cd pkg/pkg-web-multithreaded && rm README.md package.json .gitignore)
 
+.PHONY: wasm-light
```
Marcin-Radecki, line 45: Nit: we can have `wasm-without-circuits` is it imho better reflects what this target is about that `light` general name

Marcin-Radecki (approved):

crates/shielder_bindings/Makefile
kroist: The goal is to provide a separate cryptography engine. For that, we would create a new npm module, where full WASM build won't be used. That's why I'm creating a standalone module here, which would be used in the npm module

crates/shielder_bindings/src/lib.rs
kroist: We have, as the feature is included in the full WASM generation. but we don't have build without the feature

crates/shielder_bindings/Makefile
kroist: renamed

PR #254, opened 2025-06-27
BAC-177: Prover service skeleton

obrok:


tee/nix/flake.nix
```diff
@@ -28,7 +28,7 @@
 
       zkOS-monorepo-source = builtins.fetchGit {
         url = "https://github.com/Cardinal-Cryptography/zkOS-monorepo.git";
-        rev = "da70131ba57ceb678a890c1b6e553710773686a8";
```
Marcin-Radecki, line 31: Is it possible to extract this SHA to some variable that will be passed as command line argument in this script? And then call `nix  build -- abcdef1234567890 ` or, pass sha via env variable. Ideally, the SHA won't be required at all, ie once `git clone git@github.com:Cardinal-Cryptography/zkOS-monorepo.git` and nix will use already cloned sources. If this is not possible, then I'd advise to add another Makefile target in which it passes this SHA as env = `get rev-parse HEAD`.

tee/nix/flake.nix
```diff
@@ -37,6 +37,7 @@
       packages = {
         all = pkgs.linkFarm "all" [
           { name = "rewardTEE"; path = packages.rewardTEE; }
```
Marcin-Radecki, line 39: What do we do with the other set of creates? If we don't need them, then we can just use them as baseline, instead of spawning three more crates

tee/Cargo.toml
```diff
@@ -21,6 +21,7 @@ log = "0.4.27"
 sea-orm = "1.1.11"
 serde = "1.0.219"
 serde_json = "1.0.140"
+shielder-prover-common = { path = "crates/shielder-prover-common" }
```
Marcin-Radecki, line 24: Not related to this particular line, but I am wondering if we should have different repository structure, ie we have `zkOS-monorepo/crates` and it is there when we should put all Rust crates. This is because we want to benefit from the common `Cargo.lock` file, and also have reproducible builds kind-of (use the same dependencies in every production binary). So the alternative idea is e.g. to have
```
zkOS-monorepo/
   crates/
       shielder-prover-server-host/
       shielder-prover-server-tee/
       shielder-prover-server-common/
```
and then leave `zkOS-monorepo/tee` folder with current contents, that's a PoC solution for different project. In this way, we  know that what's in `crates/` is production-ready code. Wdyt?

Marcin-Radecki (approved):

tee/nix/flake.nix
obrok: I don't know! I'll find out, but I'd propose moving that problem to another PR - no reason to hold your work until I figure that out.

Marcin-Radecki: Created https://cardinal-cryptography.atlassian.net/browse/BAC-185

tee/Cargo.toml
Marcin-Radecki: Superb, makes sense, created https://cardinal-cryptography.atlassian.net/browse/BAC-186

tee/nix/flake.nix
Marcin-Radecki: Scope of https://cardinal-cryptography.atlassian.net/browse/BAC-186

PR #259, opened 2025-07-02
Removed not needed crates in TEE folder.

Marcin-Radecki:


tee/Cargo.toml
```diff
@@ -19,21 +18,20 @@ env_logger = "0.11.8"
...
 
+shielder-prover-common = { path = "crates/shielder-prover-common" }
+
 shielder-circuits = { git = "https://github.com/Cardinal-Cryptography/zkOS-circuits", rev = "df31437" }
 
 # This is not ideal - one workspace depends on another, but as long as it reproducibly produces
 # the same measurements, we are good
-type-conversions = { path = "../crates/type-conversions" }
-powers-of-tau = { path = "../crates/powers-of-tau" }
+type-conversions = { git = "https://github.com/Cardinal-Cryptography/zkOS-monorepo", rev = "4ff7e5a" }
```
Marcin-Radecki, line 36: Those two lines actually might be reverted, but that does not solve the problem of `nix build` failure

PR #261, opened 2025-07-03
Aws nitro attestation verification

kroist:
todo: add tests with real-world aws attestations


Marcin-Radecki (approved):
Looks legit! Very nice. I cross-checked this with https://github.com/aws/aws-nitro-enclaves-nsm-api/blob/main/docs/attestation_process.md

Just one general comment: we can add appropriate target to https://github.com/Cardinal-Cryptography/zkOS-monorepo/blob/main/ts/package.json#L10 so `shielder-sdk-crypto-wasm-light` is build in CI on PRs

PR #263, opened 2025-07-04
BAC-189: Support for protocol fees and memo in rust crates

JanKuczma:


contracts/Shielder.sol
```diff
@@ -724,10 +712,8 @@ contract Shielder is
 
         if (!success) revert WithdrawVerificationFailed();
 
-        uint256 newNoteIndex = _addNote(newNote);
+        newNoteIndex = _addNote(newNote);
         _registerNullifier(oldNullifierHash);
-
-        return newNoteIndex;
     }
```
kroist, line 717: return here as well?

crates/shielder-setup/lib.rs
```diff
@@ -71,12 +71,88 @@ pub mod version {
...
+    }
+}
+
+pub mod protocol_fee {
+    use alloy_primitives::{U256, U512};
+
+    /// Compute protocol fee from an amount that already includes the protocol fee
+    pub fn compute_protocol_fee_from_gross(amount: U256, protocol_fee_bps: U256) -> U256 {
+        let r = (U512::from(amount) * U512::from(protocol_fee_bps) + (U512::from(10000 - 1)))
+            / U512::from(10000);
```
kroist, line 90: what's `10000`? introduce a const for it

crates/shielder-account/src/call_data.rs
```diff
@@ -135,11 +141,18 @@ impl CallType for NewAccountCallType {
...
+            memo: extra.memo.clone(),
+        }
+        .commitment_hash();
+
         NewAccountProverKnowledge {
             id: u256_to_field(account.id),
             nullifier: u256_to_field(account.next_nullifier()),
-            initial_deposit: u256_to_field(amount),
-            caller_address: address_to_field(extra.caller_address),
+            initial_deposit: u256_to_field(amount - extra.protocol_fee),
```
kroist, line 154: protocol fee would be deducted in contract, won't it?

crates/shielder-account/src/call_data.rs
```diff
@@ -255,14 +274,21 @@ impl CallType for DepositCallType {
...
         DepositProverKnowledge {
             id: u256_to_field(account.id),
             nullifier_old: u256_to_field(nullifier_old),
             account_old_balance: u256_to_field(account.shielded_amount),
-            caller_address: address_to_field(extra.caller_address),
+            commitment: u256_to_field(commitment),
             token_address: address_to_field(token.address()),
             path: map_path_to_field(extra.merkle_path),
-            deposit_value: u256_to_field(amount),
+            deposit_value: u256_to_field(amount - extra.protocol_fee),
```
kroist, line 291: protocol fee would be deducted in contract, won't it?

crates/shielder-account/src/call_data.rs
```diff
@@ -271,11 +297,11 @@ impl CallType for DepositCallType {
     fn prepare_call_data(
         pk: &Self::ProverKnowledge,
         proof: Vec<u8>,
-        _: &Self::Extra,
+        extra: &Self::Extra,
     ) -> Self::Calldata {
         use shielder_circuits::circuits::deposit::DepositInstance::*;
         DepositCall {
-            amount: field_to_u256(pk.deposit_value),
+            amount: field_to_u256(pk.deposit_value) + extra.protocol_fee,
```
kroist, line 304: this shouldn't happen in `prepare_call_data` layer. Addition of protocol fee should be application-specific, so that the actual transferred amount is transparent to app dev.

better move addition logic to `shielder-cli`

crates/shielder-account/src/call_data.rs
```diff
@@ -153,11 +166,11 @@ impl CallType for NewAccountCallType {
     fn prepare_call_data(
         prover_knowledge: &Self::ProverKnowledge,
         proof: Vec<u8>,
-        _: &Self::Extra,
+        extra: &Self::Extra,
     ) -> Self::Calldata {
         use shielder_circuits::circuits::new_account::NewAccountInstance::*;
         NewAccountCall {
-            amount: field_to_u256(prover_knowledge.initial_deposit),
+            amount: field_to_u256(prover_knowledge.initial_deposit) + extra.protocol_fee,
```
kroist, line 173: this shouldn't happen in `prepare_call_data` layer. Addition of protocol fee should be application-specific, so that the actual transferred amount is transparent to app dev.

better move addition logic to `shielder-cli`

crates/shielder-account/src/lib.rs
```diff
@@ -68,7 +68,7 @@ impl ShielderAccount {
             ShielderAction::Deposit(data) | ShielderAction::NewAccount(data) => {
                 self.shielded_amount = self
                     .shielded_amount
-                    .checked_add(data.amount)
+                    .checked_add(data.amount - data.protocol_fee)
```
kroist, line 71: why subtract here?

kroist (changes requested):

crates/shielder-setup/lib.rs
JanKuczma: It's basis points denominator. Sure, I'll add a const for clarity.

crates/shielder-account/src/call_data.rs
JanKuczma: Yes. At the deposit (or newAccount), the user balance is increased by `amount - protocol_fee`.

crates/shielder-account/src/call_data.rs
kroist: yes, indeed if we're going to make a deposit of `amount` we need to pay `amount + protocol_fee`. 

but let's not overload the networking layer with protocol-specific details (fee math being a protocol detail)

from the perspective of contract API user, unfamiliar with function internals, it's more intuitive to think of `amount` as:
- ✅ "how much USDC do I send to the contract"
- ❌ and not "how much USDC do I want to deposit to shielder account"

On higher (application) levels, you actually might implement the feature as "user wants to shield 50 USDC, this would cost them 5 USDC premium", and in background cli actually sends 55 USDC to a contract.

JanKuczma: > I see it's already added in the shielder-cli? https://github.com/Cardinal-Cryptography/zkOS-monorepo/pull/263/files#diff-d918dcbfa38542aef8166b3c6c5dd6b900fc38c6ee3eb52c29b6b94fdc017161R49

Yes, that's an example. Note that `amount` value that we pass to `prepare_call(...)` is 100%, and the `protocol_fee` is part of it.

crates/shielder-setup/lib.rs
```diff
@@ -71,12 +71,88 @@ pub mod version {
...
 
-    /// The contract version. Currently set to 0.1.0
+    /// The contract version. Currently set to 1.1.1
     pub const fn contract_version() -> ContractVersion {
         ContractVersion {
-            note_version: 0,
+            note_version: 1,
             circuit_version: 1,
-            patch_version: 0,
+            patch_version: 1,
```
DamianStraszak, line 79: This is a little counter-intuitive, but the new version should actually be just `0.1.1` because circuits don't change, only the contract. 

There is an ADR for that https://www.notion.so/cardinalcryptography/Updating-Shielder-Contracts-and-Common-Wallet-12faf09db1a98050a6abe671aca3e71d -- you don't have to read it though.

DamianStraszak (changes requested):
Super solid :muscle: 

Left a few small ones only.

crates/shielder-setup/lib.rs
JanKuczma: I see it now. I'll change the `note_version` back to `0`.

DamianStraszak (approved):

PR #270, opened 2025-07-10
initialize tee client in sdk

kroist:


ts/shielder-sdk-crypto-wasm-light/src/circuits/teeClient.ts
```diff
@@ -0,0 +1,50 @@
...
+  provingServicePublicKey: secp.Point | undefined;
+
+  async init(
+    provingServiceUrl: string,
+    withoutAttestation: boolean
+  ): Promise<void> {
+    this.provingServiceUrl = provingServiceUrl;
+
+    const response = await fetch(`${this.provingServiceUrl}/public_key`, {
+      method: "POST",
```
Marcin-Radecki, line 23: `/public_key` is implemented as GET in `shielder-prover-server`, I assumed GET is enough here since we don't pass any parameters.

ts/shielder-sdk-crypto-wasm-light/src/circuits/teeClient.ts
```diff
@@ -0,0 +1,50 @@
...
+    const data: TeePublicKeyResponse =
+      (await response.json()) as TeePublicKeyResponse;
+    if (!data.public_key || !data.attestation_document) {
+      throw new Error(
+        "Invalid response from TEE service: missing public key or attestation document"
+      );
+    }
+
+    if (withoutAttestation) {
+      this.provingServicePublicKey = secp.Point.fromHex(data.public_key);
```
Marcin-Radecki, line 42: Nit: just set ` this.provingServicePublicKey = secp.Point.fromHex(data.public_key);
Comment` always, and inside this `if` just `return`

kroist: that might be not a valid pubkey in case if we're expecting attestation.

Marcin-Radecki: ah, you're right

Marcin-Radecki (approved):

PR #271, opened 2025-07-10
Encrypt as well user public key

Marcin-Radecki:
Apart from a circuit payload, we need to encrypt user's public key as well.


tee/crates/shielder-prover-common/src/protocol.rs
```diff
@@ -18,7 +18,7 @@ pub enum Request {
     /// `pub_sk` is a public key generated by the user, which TEE should use to encrypt proof in the response
     GenerateProof {
         payload: Vec<u8>,
-        user_public_key: String,
+        user_public_key: Vec<u8>,
```
kroist, line 21: why not a part of the `payload`?

Marcin-Radecki:
I'm merging, despite of broken CI, as not related to this PR

PR #291, opened 2025-07-22
terraform script for tee setup on ec2

kroist:


tee/tf/main.tf
```diff
@@ -0,0 +1,263 @@
...
+echo "Downloading enclave image from ${var.eif_url}"
+curl -sSL "${var.eif_url}" -o /home/ec2-user/image.eif
+chown ec2-user:ec2-user /home/ec2-user/image.eif
+
+# Set TEE_CID
+export TEE_CID=16
+
+# Run the enclave
+echo "Starting Nitro Enclave with CID=$${TEE_CID}"
+nitro-cli run-enclave --cpu-count 2 --memory 2048 --enclave-cid $${TEE_CID} --eif-path /home/ec2-user/image.eif --debug-mode
```
Marcin-Radecki, line 174: Are we going to add some flag/variable for `--debug-mode`? For production, we should not be using `--debug-mode`

kroist: added a tf variable

